PR #13, opened 2019-06-29
chore: tslint to eslint (fixes #12)

### LINKED ISSUE #12: tslint => eslint
refs: https://eslint.org/blog/2019/01/future-typescript-eslint

aladdin-add:


.eslintrc.js
```diff
@@ -0,0 +1,31 @@
...
+      rules: {
+        '@typescript-eslint/no-use-before-define': [2, { functions: false }], // https://github.com/eslint/eslint/issues/11903
+        '@typescript-eslint/indent': 0,
+
+        // TODO: enable it when all problems addressed
+        '@typescript-eslint/explicit-function-return-type': 0,
+        '@typescript-eslint/no-explicit-any': 0,
+        '@typescript-eslint/class-name-casing': 0,
+        '@typescript-eslint/camelcase': 0
+      }
```
aladdin-add: hopefully these rules can be enabled later.

aladdin-add:
@KFlash there is something you might have to address :)
https://circleci.com/gh/meriyah/meriyah/1779?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

KFlash:
@aladdin-add I looked into it, and many of this "errors" happens because of bugs within ESLint, so you may fix the bugs first?

Example `let { token, tokenValue, tokenIndex, linePos, colPos } = parser;` Here you got this error

` 2403:11  error  'token' is never reassigned. Use 'const' instead                               prefer-const`

The funny thing is that `tokenValue` is re-assigned 3 lines down, so the block can't be a constant.

There are other bugs in ESLint as well. I think you will find them :)

futagoza:
@KFlash Use this:
```json
"prefer-const": [ "error", { "destructuring": "all" } ],
```
Setting the `destructing` option to `all` will ensure the rule checks that all variables are not re-assigned before ESLint throws an error like that. See [https://eslint.org/docs/rules/prefer-const#destructuring](https://eslint.org/docs/rules/prefer-const#destructuring)

aladdin-add:
> You also have this
> ```bash
> /home/circleci/repo/test/parser/lexical/lexical.ts
>    2:10  warning  'pass' is defined but never used          @typescript-eslint/no-unused-vars
>   22:8   error    Unexpected var, use let or const instead  no-var
>   24:10  error    Unexpected var, use let or const instead  no-var
>   69:10  error    Unexpected var, use let or const instead  no-var
> ```
> The code is this
> ```js
> for (var k = 0; k < letbinds.length; ++k) {}
> ```
> If you change it to a const as suggested you get multiple errors. Changing it to let would work.

I think this is working as expected, the rule is to disallow the usage of vars, which is discouraged in ES6.

the autofixing is: `var` => `let`(by no-vars) => `const`(by prefer-const). not sure why it is not autofixed in this case.

package.json
```diff
@@ -64,11 +64,13 @@
     "@types/node": "^12.0.10",
     "@typescript-eslint/eslint-plugin": "^1.11.0",
     "@typescript-eslint/parser": "^1.11.0",
+    "chalk": "^2.4.2",
     "coveralls": "^3.0.4",
     "cross-env": "^5.2.0",
     "eslint": "^6.0.1",
     "eslint-plugin-import": "^2.18.0",
     "eslint-plugin-node": "^9.1.0",
+    "glob": "^7.1.4",
```
KFlash: glob is for?

aladdin-add: https://github.com/meriyah/meriyah/blob/7e8d0463a13d019d16fe4ec2906801e41edea36d/scripts/prettier.js#L13

aladdin-add:
emm.. it's strange when I changed `var` to `let`, got some errors like this:
image

it should not happen in my knowledge, maybe something related to ts-node?

KFlash:
You are not the first one to get that error :) Should be impossible yes. Dunno what cause this, but let me try to hard-write this tests locale

KFlash:
@aladdin-add The issue you had should have been solved. I just removed the tests.

KFlash:
@aladdin-add I removed the tests because many of them or similar tests exist if you scroll down in that file. Some of the other tests are almost identical tests that you find in `/lexical/block.ts`.

There are no points to test the same things twice :)

aladdin-add:
I thinks there is a bug here: when using var, the tested code is actually `{undefined}undefined` -- that's why `t.doesNotThrow()` works.
image

anyway, no `var`, please use `let/const`! :)
esp. when using callbacks.

KFlash:
I noticed it. Nothing to do with the parser itself. If you use a `console.log` it can not find the values in the array and uses  `undefined` instead. However the last `for loop` find the elements in the array correctly.

KFlash:
@aladdin-add You was right. It was a bug and I think it's related to Mocha. At least I solved it with a workaround :)

KFlash:
@aladdin-add This PR now fails because of `403 Forbidden: estraverse@4.2.0`.

aladdin-add:
@KFlash fixed~

KFlash:
@aladdin-add I approve this, but you should take a look at the lint issues above before merging.

aladdin-add:
will you fix the type-related issue?

KFlash:
I'm not on a computer today, and I'm not so familiar with all the ESLint rules. Could you do it?

KFlash:
I merged this now. Fixed a few issues and skipped a few.

Issue #15, opened 2019-07-03
Implement JSX syntax parsing


KFlash:
This issue ticket exist just to keep track of JSX implementation.

This is WIP at the moment, and a new option - `jsx` - will be added to allow JSX syntax parsing.

KFlash:
@aladdin-add Any ideas about JSX annotations? How should we handle them?
`/** @jsx CUSTOM_DOM */<a></a>`

PR #16, opened 2019-07-05
feat(parser): WIP: Enable JSX parsing

KFlash:
Adds support for JSX syntax parsing. All available JSX features have been implemented, and I have also made sure we are not suffering from the known issues in Acorn JSX tracker and same for Babel.

Fixes https://github.com/meriyah/meriyah/issues/15


KFlash:
@aladdin-add @nchanged This is still WIP. I need to fix a few non-failing tests, optimize it's performance and make sure loc tracking is correct before merge to Master.

Can anyone test this and give feedback?

KFlash:
Optimization and simplification is the last thing to fix. This will make the last non-failing cases to fail as expected and then this can be merged to Master.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "meriyah",
-  "version": "1.2.6",
+  "version": "1.3.0",
```
aladdin-add: it should be changed when you want to do a release. :)

src/common.ts
```diff
@@ -409,3 +409,30 @@ export function finishNode<T extends Node>(
...
+    }
+
+    if (lhs.type === 'JSXMemberExpression') {
+      (
+        isEqualTagName(lhs.object, rhs.object) + '.' +
+        isEqualTagName(lhs.property, rhs.property)
+    ) === (
+      isEqualTagName(lhs.object, rhs.object) + '.' +
+        isEqualTagName(lhs.property, rhs.property)
+  )
```
aladdin-add: seems the (...) === (...) is useless, am I missing something?

KFlash:
@aladdin-add You didn't see the fragment tests? They [exist here](https://github.com/meriyah/meriyah/blob/jsx/test/parser/miscellaneous/jsx.ts#L2011). I have implemented fragments :)

Regarding extensibility. How would you do it? Any ideas? We could offer callbacks, but would that help?

src/common.ts
KFlash: That code is useless at the moment, so you are right. I need to fix it and the remaining optimizing.

KFlash:
@aladdin-add I'm merging this now and will improve performance in general all together. JSX should work as expected.

src/common.ts
KFlash: @aladdin-add There are no way to extend Meriyah like you suggested. This parser is build for performance. However. I invented you to [buntis](https://github.com/buntis/buntis). 

That will be a JS & TS parser that already now accept plugins. And ofc. allow you to parse TypeScript.

Hope that solves a few issues you may have :)

PR #26, opened 2019-08-02
WIP! Implements nullish coalescing and optional chaining (stage 3)

KFlash:
This implements support for the latest Stage 3 features.

- [x] Nullish coalescin

- [ ] optional chaining

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@
 * [Decorators](https://github.com/tc39/proposal-decorators)
 * [Class Public Instance Fields & Private Instance Fields](https://github.com/tc39/proposal-class-fields)
 * [Hashbang grammar](https://github.com/tc39/proposal-hashbang)
+* [Nullish coalescing Operator](https://github.com/tc39/proposal-nullish-coalescing)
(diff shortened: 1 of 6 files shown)


KFlash:
I'm still experimenting with the binary expr validationfor nullish coalescing to improve performance. And try my best to avoid same logic as Babel :) For now I'm only passing the token as func argument, but I think there exist a better solution...

```ts
if (op === tt.nullishCoalescing) {
          if (
            left.type === "LogicalExpression" &&
            left.operator !== "??" &&
            !(left.extra && left.extra.parenthesized)
          ) {
            throw this.raise(
              left.start,
              `Nullish coalescing operator(??) requires parens when mixing with logical operators`,
            );
          } else if (
            node.right.type === "LogicalExpression" &&
            node.right.operator !== "??" &&
            !(node.right.extra && node.right.extra.parenthesized)
          ) {
            throw this.raise(
              node.right.start,
              `Nullish coalescing operator(??) requires parens when mixing with logical operators`,
            );
          }
        }
```

KFlash:
Merging this as it is. Optional chaining will be separate PR due to missing ESTree info.

PR #28, opened 2019-08-02
WIP! Implement optional chaining and import.meta

KFlash:
This is still a WIP, and can't be finished before we know the ESTree specs for it.

 For now this implementation is based on this [ESTree discussion](https://github.com/estree/estree/issues/146) and how optional chaining is done in V8, SM and Webkit.

Import.meta added here to avoid conflicts later on.


src/parser.ts
```diff
@@ -3621,7 +3621,8 @@ export function parseMemberOrUpdateExpression(
   inGroup: 0 | 1,
   start: number,
   line: number,
-  column: number
+  column: number,
+  chained: 0 | 1 = 0
```
KFlash: @aladdin-add `chained: 0 | 1 = 0` is temporary because we don't know the ESTree specs yet. Added to in line with Babel parser :)  When you figure it out with ESTree this has to either 1) be removed or 2) moved up to be in sync with`inGroup` and `= 0` removed so every call to `parseMemberOrUpdateExpression` has to have a `0` as func arg.

Example: `parseMemberOrUpdateExpression(parser, context, /* inGroup */ 0, /* chained */ 0, start, line, column)`

If you figure out that the `OptionalMemberExpression` shouldn't be there, well just remove it and use the original one and update `estree.ts` as well.

KFlash:
@aladdin-add After [studying the specs](https://tc39.es/proposal-optional-chaining/) I think Babel's implementation is wrong and not how it will be. 

My current changes are more in line with what suggested in one of the ESTree discussion from 2017.

I will push this changes soon, but here is an example.

**Note** the new `OptionalChain AST node`

You parse this:  `a.b?.c`

**Outputs:**

```ts
{
          "body": [
            {
              "expression": {
                "expr": {
                  "computed": false
                  "object": {
                    "computed": false
                    "object": {
                      "name": "a"
                      "type": "Identifier"
                    }
                    "property": {
                      "name": "b"
                      "type": "Identifier"
                    }
                    "type": "MemberExpression"
                  }
                  "optional": true
                  "property": {
                    "name": "c"
                    "type": "Identifier"
                  }
                  "type": "MemberExpression"
                }
                "type": "OptionalChain"
              }
              "type": "ExpressionStatement"
            }
          ]
          "sourceType": "script"
          "type": "Program"
        }
```

KFlash:
@aladdin-add and others. I enabled optional chaining and other new Stage 3 proposals available  in the REPL so this PR can be tested live there. 

REPL:  https://meriyah.github.io/meriyah/

KFlash:
Fixed the performance regression. There was a regression while parsing larger libraries. Hard to see for people that doesn't know this code :)

Here is the new benchmark results. And we are good to go 👍 

perf1

KFlash:
@aladdin-add Added `import.meta` into this PR to avoid conflicts. Any progress with the ESTree specs?

PR #32, opened 2019-08-08
WIP: Context based escape keyword validation

KFlash:
This PR fixes all escaped keywords. I ripped out all escape keyword validations and replaced it with a context based solution. 

A context based solution means we never need to do a `if / else` etcetera. Instead this is caught "the normal way" in the places we are either 1) expecting a keyword or 2) consumes there should be a keyword.

For the `static keyword` edge cases - I'm passing out the escaped keyword as an normal identifier.

Benefits of all this is ofc. performance :)

**Note:** Error messages will change to 

- `Unexpected token: 'escaped keyword'" `

- `Unexpected token: 'escaped future reserved keyword'"`

Only 2 non-failing edgy cases left to fix before cleaning up this mess :)


src/errors.ts
```diff
@@ -287,8 +287,8 @@ export const errorMessages: {
   [Errors.IllegalReturn]: 'Illegal return statement',
   [Errors.InvalidForLHSBinding]: 'The for-header left hand side binding declaration is not destructible',
   [Errors.InvalidNewTarget]: 'new.target only allowed within functions',
-  [Errors.InvalidEscapedKeyword]: 'Keywords must be written literally, without embedded escapes',
-  [Errors.MissingPrivateName]: "''#' not followed by identifier",
+  [Errors.InvalidEscapedKeyword]: 'Kasdfafdsadfsy, without embedded escapes',
```
aladdin-add: the message looks confusing. is it expected?

KFlash: `[Errors.InvalidEscapedKeyword` will be gone and the `Errors.MissingPrivateName` seems to have an typo - an extra `'`. 

What's confusing? Everything can be changed :)

aladdin-add: what's the meaning of `Kasdfafdsadfsy`

KFlash:
@aladdin-add In general escaped keywords are on the slow path anyways, so I don't care too much about the validation the lexer has to do for this to work. But I'm using bit masks so shouldn't loose any performance even on the slow path.

The benefits with this approach is that we never need to validate against any escaped keyword or care about them at all, and it will improve overall performance.

What's is a real pain in the ass is to implement this correctly, and I haven't seen this been done before so I'm working blindfolded :)

src/errors.ts
KFlash: Just a typo :) It's all fixed

PR #49, opened 2019-10-03
Fix tracking of parser location regardless of line-ending format

lawrence-dol:
Abstract the handling of a line break to be agnostic of whether the sequence is any of CR,CRLF, LF, or LFCR.

### DIFF
--- a/src/lexer/comments.ts
+++ b/src/lexer/comments.ts
@@ -67,7 +67,6 @@ export function skipMultiLineComment(parser: ParserState, state: LexerState): Le
   const { index } = parser;
   while (parser.index < parser.end) {
     while (parser.currentChar === Chars.Asterisk) {
-      state &= ~LexerState.LastIsCR;
       if (advanceChar(parser) === Chars.Slash) {
         advanceChar(parser);
         if (parser.onComment)
--- a/src/lexer/common.ts
+++ b/src/lexer/common.ts
@@ -6,7 +6,6 @@ import { report, Errors } from '../errors';
 export const enum LexerState {
   None = 0,
   NewLine = 1 << 0,
-  //LastIsCR = 1 << 2 // no longer care about CR's
 }
 
 export const enum NumberKind {
@@ -42,28 +41,6 @@ export function advanceChar(parser: ParserState): number {
   }
   return parser.currentChar;
 }
-//export function advanceChar(parser: ParserState): number {
-//  parser.column++;
-//  return advanceAndLawrenceDolTheCRLF(parser);
-//}
-//
-//export function advanceAndLawrenceDolTheCRLF(parser: ParserState) {
-//  parser.currentChar = parser.source.charCodeAt(++parser.index);
-//  if (parser.index < parser.end) {
-//    const cur = parser.currentChar;
-//    const nxt = parser.source.charCodeAt(parser.index + 1);
-//    if (
-//      (cur == Chars.CarriageReturn && nxt == Chars.LineFeed) ||
-//      (cur == Chars.LineFeed && nxt == Chars.CarriageReturn)
-//    ) {
-//      parser.currentChar = Chars.LineFeed;
-//      parser.index++;
-//    } else if (cur === Chars.CarriageReturn) {
-//      parser.currentChar = Chars.LineFeed;
-//    }
-//  }
-//  return parser.currentChar;
-//}
 
 /**
  * Consumes multi unit code point
@@ -94,21 +71,6 @@ export function consumeLineBreak(parser) {
   parser.line++;
   parser.currentChar = advanceChar(parser); // important to use advanceChar() so CR/LF are handled consistently; also, col was just reset to 0, so correct for advanceChar() to do column++.
 }
-//export function consumeLineFeed(parser: ParserState, state: LexerState): void {
-//  advanceAndLawrenceDolTheCRLF(parser);
-//  parser.flags |= Flags.NewLine;
-//  if ((state & LexerState.LastIsCR) === 0) {
-//    parser.column = 0;
-//    parser.line++;
-//  }
-//}
-//
-//export function consumeLineBreak(parser: ParserState): void {
-//  parser.flags |= Flags.NewLine;
-//  parser.column = 0;
-//  parser.line++;
-//  advanceAndLawrenceDolTheCRLF(parser);
-//}
 
 // ECMA-262 11.2 White Space
 export function isExoticECMAScriptWhitespace(code: number): boolean {
--- a/src/lexer/common.ts
+++ b/src/lexer/common.ts
@@ -65,7 +65,7 @@ export function consumeMultiUnitCodePoint(parser: ParserState, hi: number): 0 | 
 /**
  * Use to consume a line feed instead of `consumeLineBreak`.
  */
-export function consumeLineBreak(parser) {
+export function consumeLineBreak(parser: ParserState): void {
   parser.flags |= Flags.NewLine;
   parser.column = 0;
   parser.line++;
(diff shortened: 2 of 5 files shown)


src/lexer/common.ts
```diff
@@ -28,26 +28,42 @@ export const enum NumberKind {
...
-      parser.currentChar = Chars.LineFeed;
-      parser.index++;
-    } else if (cur === Chars.CarriageReturn) {
-      parser.currentChar = Chars.LineFeed;
-    }
+
+  if(parser.index < parser.end) {
+      const cur = parser.currentChar;
+      const nxt = parser.source.charCodeAt(parser.index + 1);
+      if((cur == Chars.CarriageReturn && nxt == Chars.LineFeed) || (cur == Chars.LineFeed && nxt == Chars.CarriageReturn)) {
```
KFlash: Can't you use a second func arg that validates to true if CR. That way you could simplify and optimize this line?

src/lexer/common.ts
```diff
@@ -28,26 +28,42 @@ export const enum NumberKind {
...
+//      (cur == Chars.LineFeed && nxt == Chars.CarriageReturn)
+//    ) {
+//      parser.currentChar = Chars.LineFeed;
+//      parser.index++;
+//    } else if (cur === Chars.CarriageReturn) {
+//      parser.currentChar = Chars.LineFeed;
+//    }
+//  }
+//  return parser.currentChar;
+//}
```
KFlash: Just remove the code you commented out to make the source clean

src/lexer/common.ts
```diff
@@ -72,21 +88,27 @@ export function consumeMultiUnitCodePoint(parser: ParserState, hi: number): 0 |
...
-  if ((state & LexerState.LastIsCR) === 0) {
-    parser.column = 0;
-    parser.line++;
-  }
-}
-
-export function consumeLineBreak(parser: ParserState): void {
-  parser.flags |= Flags.NewLine;
+export function consumeLineBreak(parser) {
+  parser.flags |= 1;
```
KFlash: Avoid using number here. Use the const enum value for readability. The compiler will fix the number part

src/lexer/common.ts
```diff
@@ -72,21 +88,27 @@ export function consumeMultiUnitCodePoint(parser: ParserState, hi: number): 0 |
...
+//    parser.line++;
+//  }
+//}
+//
+//export function consumeLineBreak(parser: ParserState): void {
+//  parser.flags |= Flags.NewLine;
+//  parser.column = 0;
+//  parser.line++;
+//  advanceAndLawrenceDolTheCRLF(parser);
+//}
```
KFlash: Same as earlier. Just remove the code that have been commented out

src/lexer/scan.ts
```diff
@@ -534,14 +534,20 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
...
-          break;
+        // Now we only handle "line breaks", regardless of whether indicated by CR, CRLF, LF, or LFCR.
+        //case Token.CarriageReturn:
+        //  consumeLineBreak(parser);
+        //  state |= LexerState.NewLine;
+        //  break;
+        //case Token.LineFeed:
+        //  consumeLineFeed(parser, state);
+        //  state = (state & ~LexerState.LastIsCR) | LexerState.NewLine;
+        //  break;
```
KFlash: Same as earlier. Remove the code that have been commented out.

src/lexer/scan.ts
```diff
@@ -534,14 +534,20 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
           break;
 
         case Token.CarriageReturn:
+        case Token.LineFeed:
           consumeLineBreak(parser);
```
KFlash: Regarding my comment earlier. If you have a 2nd func arg you can pass it as true for CR. That was the entire point with the bitmasks you removed :)

src/lexer/common.ts
KFlash: And you should always use `===` and not `==`

src/lexer/comments.ts
```diff
@@ -67,7 +67,6 @@ export function skipMultiLineComment(parser: ParserState, state: LexerState): Le
   const { index } = parser;
   while (parser.index < parser.end) {
     while (parser.currentChar === Chars.Asterisk) {
-      state &= ~LexerState.LastIsCR;
```
KFlash: This entire single line comment can be rewritten and optimized for this purpose I guess. Look at this pseudo code

```ts
let lastIsCR = 0

while (this.hasNext()) {
    switch (this.current()) {
    case "\r": lastIsCR = 2 // So it gets decremented to 1
    case "\n": case "\u2028": case "\u2029":
        if (!--lastIsCR) this.line++
        this.column = 0
        this.index++
        this.hasLineTerminator = true
        return // end loop
    default:
        if (lastIsCR) { this.line++; lastIsCR = 0 }
        this.column++
        this.index++
    }
}
```

src/lexer/scan.ts
lawrence-dol: There's no need to do anything different for CR or LF. The two token values can probably be coalesced into `Token.LineBreak`.

src/lexer/comments.ts
lawrence-dol: This is no need for specific handling of CR/LF anywhere other than `advanceChar`, and there it need only treat `CR`, `CRLF`, `LFCR` as a line `LF`. Similarly, the only place column is reset, and line incremented should be in `consumeLineFeed`.

src/lexer/common.ts
lawrence-dol: Yeah, I wasn't expect to be do a pull request so I left in the altered code to ease your checking.

src/lexer/common.ts
lawrence-dol: This is testing for a `CRLF` or `LFCR` sequence, not just CR. Starting from the beginning any LF, CR, CRLF, or LFCR must be treated as a line-break. This logic works for a file which is entirely one of these, or a mix. 

Now, in practice LFCR was never used by any OS though it's technically allowed instead of CRLF, lone CR was used only by Mac OS9 and earlier, and today only LF, and CRLF really exist in practice. Depending on git settings and editors it's quite possible for an editor to emit a mixed LF & CRLF file. This logic copes with all combinations, and is simple enough.

To cope with only LF and CRLF (or a mix of them), the above check for `(cur === LF && nxt === CR)` could be dropped.

I don't see a subfunction being sufficiently clearer to warrant the one-liner, but if you want I will:

    function isLineBreakPair(cur,nxt) {
        return (cur == Chars.CarriageReturn && nxt == Chars.LineFeed) || (cur == Chars.LineFeed && nxt == Chars.CarriageReturn);
    }

src/lexer/common.ts
lawrence-dol: I started off working on the generated JS, so the constants had been eliminated. This was an oversight.

src/lexer/comments.ts
lawrence-dol: Incidentally, I couldn't find a constant for the `8` and `8232` bit-flags used `comments.ts`.

src/lexer/common.ts
lawrence-dol: OK, I did that.

src/lexer/scan.ts
lawrence-dol: See my final push; I've combined these two to `Token.LineBreak`.

lawrence-dol:
@KFlash : I think that covers it.

src/lexer/scan.ts
KFlash: Seems good now. I'm going to merge this and ship  a new version

lawrence-dol:
BTW, I don't understand why LGTM has flagged an unused import in `common.ts`; the imported names are both used in the module.

KFlash:
@lawrence-dol Dunno either, but I need to revert this merge for now

KFlash:
@lawrence-dol See open PR now and try to address the issues before merging again.

PR #53, opened 2019-10-20
Fix html comments

GuyLewin:
HTML comments include parts of their prefix

### DIFF
--- a/src/lexer/scan.ts
+++ b/src/lexer/scan.ts
@@ -232,13 +232,7 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
             }
             if (ch === Chars.Exclamation) {
               // Treat HTML begin-comment as comment-till-end-of-line.
-              if (
-                parser.source.charCodeAt(parser.index + 1) === Chars.Hyphen &&
-                parser.source.charCodeAt(parser.index + 2) === Chars.Hyphen
-              ) {
-                // Skip !--
-                advanceChar(parser);
-                advanceChar(parser);
+              if (advanceChar(parser) === Chars.Hyphen && advanceChar(parser) === Chars.Hyphen) {
                 advanceChar(parser);
                 state = skipSingleHTMLComment(parser, state, context, CommentType.HTMLOpen);
                 continue;
@@ -355,7 +349,6 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
             advanceChar(parser);
             if ((state & LexerState.NewLine || isStartOfLine) && parser.currentChar === Chars.GreaterThan) {
               if ((context & Context.OptionsWebCompat) === 0) report(parser, Errors.HtmlCommentInWebCompat);
-              // Skip >
               advanceChar(parser);
               state = skipSingleHTMLComment(parser, state, context, CommentType.HTMLClose);
               continue;
--- a/src/lexer/scan.ts
+++ b/src/lexer/scan.ts
@@ -232,8 +232,14 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
             }
             if (ch === Chars.Exclamation) {
               // Treat HTML begin-comment as comment-till-end-of-line.
-              if (advanceChar(parser) === Chars.Hyphen && advanceChar(parser) === Chars.Hyphen) {
-                advanceChar(parser);
+              const index = parser.index + 1;
+              if (
+                index + 1 < parser.source.length &&
+                parser.source.charCodeAt(index) === Chars.Hyphen &&
+                parser.source.charCodeAt(index + 1) == Chars.Hyphen
+              ) {
+                parser.column += 3;
+                parser.currentChar = parser.source.charCodeAt((parser.index += 3));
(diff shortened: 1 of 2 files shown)


src/lexer/scan.ts
```diff
@@ -233,9 +233,13 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
             if (ch === Chars.Exclamation) {
               // Treat HTML begin-comment as comment-till-end-of-line.
               if (
-                parser.source.charCodeAt(parser.index + 2) === Chars.Hyphen &&
```
KFlash: This code is an optimization tweak. Checking for the 2nd 'Chars.Hypen' first avoids extra check for the first  'Chars.Hyphen' if 2nd doesn't exist. Kind of micro-optimization.

src/lexer/scan.ts
```diff
@@ -233,9 +233,13 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
...
-                parser.source.charCodeAt(parser.index + 1) === Chars.Hyphen
+                parser.source.charCodeAt(parser.index + 1) === Chars.Hyphen &&
+                parser.source.charCodeAt(parser.index + 2) === Chars.Hyphen
               ) {
+                // Skip !--
+                advanceChar(parser);
```
KFlash: This could be tweaked to avoid member access and multiple `parser.source.charCodeAt` calls. E.g  `parser.source.charCodeAt(parser.index +3)`

src/lexer/scan.ts
GuyLewin: Ooh I like it, but I was worried from the edge case that `parser.index + 1 == EOF` and there's no `parser.index + 2`. I guess NaN will return which is kind of OK, but I wanted to avoid that. Should we maybe check that `pos` doesn't exceed `parser.end + 2`?

KFlash: Better yes to check for `parser.end + 2`.

src/lexer/scan.ts
GuyLewin: Not sure I follow, I want to make `currentChar` point to the beginning of the comment value so skipSingleLineComment will read the value correctly.

KFlash: This should work the same way `parser.currentChar = parser.source.charCodeAt(parser.index +3)` minus the extra `parser.source.charCodeAt` you use when using `advanceChar`

GuyLewin: Yeah but you're already reading charCodeAt when testing for ```parser.source.charCodeAt(parser.index + 2) === Chars.Hyphen &&
parser.source.charCodeAt(parser.index + 1) === Chars.Hyphen```
I suggest we convert that code to use advanceChar twice since we really do want to advance column index and check the current char.
The alternative would be to support something like:
```
export function advanceChar(parser: ParserState, advanceAmount: number = 1): number {
  parser.column += advanceAmount;
  return (parser.currentChar = parser.source.charCodeAt(parser.index += advanceAmount));
}
```
and call advanceChar(parser, 3)

KFlash: Agree there, so it's better to convert the 2 first lines to use advanceChar.

KFlash:
@GuyLewin Should I merge this as it is?

GuyLewin:
@KFlash no not yet, I'm not home right now but I'll take care of it soon

KFlash:
okay.

KFlash:
Off topic, but any ideas why the column value is off by 1 in a few cases? I compared it earlier with my private code and other parsers. E.g. `x={ dsda(){ return (a=super.foo) => a; }}`. The `start` and `end` values are correct, but the start column value is wrong. Similar cases related to super property.

I need to investigate this...

GuyLewin:
@KFlash I'll have a look

GuyLewin:
@KFlash I pushed the advanceChar fix and I think it's good now. I know I can push stuff by my own but I'll give you a day or two if you happen to look at it before 😄

KFlash:
This looks good! But note the new LGTM alert - "Identical operands".

GuyLewin:
@KFlash yeah, hold on I'll try something else

KFlash:
Is this ready to be merged?

GuyLewin:
@KFlash yeah

KFlash:
Ok. Can you merge it? And I will publish a new version soon as I get time

GuyLewin:
@KFlash yeah, I still need 1 reviewer approval 😄

KFlash:
It's approved now

GuyLewin:
Merged, thanks

KFlash:
@GuyLewin If I publish some private code. Do you think you have time to "merge" some of the code into Meriyah for a 2.0 release? It will solve some known issues

GuyLewin:
@KFlash yeah, it's the weekend over here so I have some time. I wanted to ask you some questions about the strategy of meriyah, maybe we can email about it

KFlash:
Yeah sure. Just send an email to "klechon123 at gmail.com."

KFlash:
@GuyLewin Did you get the code?

PR #54, opened 2019-10-26
Performance upgrades

GuyLewin:


GuyLewin:
Thanks! I saw your work on 2.0, it’s awesome. It’ll take me some time to port

KFlash:
Yeah in current 2.0 I focused on tweaking the scanner code. Note that a local boolean turns out to be faster. You see how I use it in that source.

KFlash:
@GuyLewin Just noticed it. See [this lines](https://github.com/GuyLewin/meriyah/blob/2c003d4ca82b1f551da9bf3c508b3af2717026c3/src/lexer/comments.ts#L109-L115) . You can combine this into one single "else" and only modify the state once.

GuyLewin:
@KFlash yeah but then in case `parser.currentChar ^ Chars.LineSeparator`- we're setting `state` twice. Right now we're setting it once but the code is semi-duplicate.

KFlash:
Any possibility we could ship the `parser.currentChar` and set the value right before return? In the 2.0 code I'm not caching anything. Look at the multiline comment code there. I improved it even further.

PR #56, opened 2019-11-03
Allow fast testing of specific files

GuyLewin:
Until now I had to run `npm test` which takes long to complete. It compiles .ts to JS before running in Mocha.
I created another Node script to help us test specific UTs fast while developing, and used Mocha + ts-node to compile TypeScript instead of cross-env.


tsconfig.json
```diff
@@ -7,8 +7,6 @@
     "declarationMap": true,
     "forceConsistentCasingInFileNames": true,
     "importHelpers": true,
-    "lib": ["esnext"],
```
GuyLewin: ts-node failed with 'esnext'. Why do we compile to esnext? I removed it because I don't think it's necessary but let me know.

Issue #60, opened 2019-11-21
In Object pattern can we create 2 different objects for key and value


peeyush31:
if the key is not specified then the value and left of assignment pattern is the same object
```js
// 1st case
var {x} = {}

//2nd case
var {y=10} = {}
```
In these cases the key is not defined:
 - 1st case the `key` and `value` are the same object
 - 2nd case the `key` and `value.left` the left part of assignmentPattern are the same object

As a consumer to the parser I have to add properties to the AST nodes as the nodes are same object they properties get automatically added.

It will be good if the parser supports this and some  other people might face a similar issue.

peeyush31:
@KFlash I can create a PR if you are OK with this change

KFlash:
Hi. If you want you can do an PR and add this changes as an option. As it is now Meriyah are following the ESTree specs, so changes like this has to be behind and option.

KFlash:
@GuyLewin Do you want to look into this?

aladdin-add:
closing, as it seems has been resolved. thanks!

PR #61, opened 2019-12-19
Create unique key and value for Object pattern

peeyush31:
this will fix the issue https://github.com/meriyah/meriyah/issues/60


KFlash:
Sorry late reply. Internet issues in my end, but merged!

PR #67, opened 2020-01-25
Create SECURITY.md

leejeongro:


KFlash:
cc/ @aladdin-add

SECURITY.md
```diff
@@ -0,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
```
aladdin-add: I don't think this is needed, as we didn't maintain multi-branch releases.

SECURITY.md
```diff
@@ -0,0 +1,21 @@
...
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.1.x   | :white_check_mark: |
+| 5.0.x   | :x:                |
+| 4.0.x   | :white_check_mark: |
+| < 4.0   | :x:                |
+
+## Reporting a Vulnerability
+
+Use this section to tell people how to report a vulnerability.
```
aladdin-add: we do not suggest to use GitHub for filing security issues. Doing so is insecure and may put users at risk.

leaving an email address sounds good to me.  @KFlash and/or @aladdin-add(weiran.zsd@outlook.com)

PR #75, opened 2020-03-31
Change key order of other nodes

eobodo:
I checked how both Acorn orders node keys and how Estraverse walks the tree.
The following nodes have been adjusted to make dynamic tree iterators act similarly:

1. ArrowFunctionExpression
2. ExportNamedDeclaration
3. ForStatement
4. ForOfStatement
5. FunctionDeclaration
6. FunctionExpression

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2300,9 +2300,9 @@ export function parseForStatement(
 
       return finishNode(parser, context, start, line, column, {
         type: 'ForOfStatement',
-        body,
         left: init,
         right,
+        body,
         await: forAwait
       });
     }
@@ -2359,10 +2359,10 @@ export function parseForStatement(
 
   return finishNode(parser, context, start, line, column, {
     type: 'ForStatement',
-    body,
     init,
     test,
-    update
+    update,
+    body
   });
 }
 
@@ -3085,9 +3085,9 @@ function parseExportDeclaration(
 
   return finishNode(parser, context, start, line, column, {
     type: 'ExportNamedDeclaration',
-    source,
+    declaration,
     specifiers,
-    declaration
+    source,
   });
 }
 
@@ -4840,11 +4840,11 @@ export function parseFunctionDeclaration(
 
   return finishNode(parser, context, start, line, column, {
     type: 'FunctionDeclaration',
+    id,
     params,
     body,
     async: isAsync === 1,
     generator: isGenerator === 1,
-    id
   });
 }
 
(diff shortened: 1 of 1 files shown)


KFlash:
Thanks for this! I will merge and publish a new version. 

Have you seen my private parser code - [seafox](https://github.com/KFlash/seafox/blob/master/src/seafox.ts)? Meriyah was build by my students, and I'm maintaining it now.

eobodo:
No problem at all. Seafox looks like a very interesting project. I remember seeing some mention of it, but didn't realize it was already open-sourced. I'll be sure to test it out soon.

KFlash:
Seafox is my own code. I'm now working on a incremental parser. Still experimenting with it

eobodo:
That's awesome. An incremental Javascript parser would be very useful. Especially in this age of large 
 (js) code bases and growing IDE usage, I can see many uses cases were an incremental parser will be handy.

leeoniya:
also check out the incremental parser built for CodeMirror6 (inspired by [tree-sitter](https://tree-sitter.github.io/tree-sitter/)) : https://marijnhaverbeke.nl/blog/lezer.html

https://github.com/lezer-parser/lezer
https://github.com/lezer-parser/lezer-generator
https://github.com/lezer-parser/lezer-tree

eobodo:
Lezer (and CodeMirror6) look awesome, thanks for sharing. Funny enough, incremental parsing within an in-browser editor was one use case I had in mind. That of course makes of using tree-sitter more difficult. Glad to see something similar prop up in the js ecosystem. I'll be sure to check it out.

KFlash:
@leeoniya @eobodo `lezer` seems very complex, and what about the performance vs other parsers?

Maybe this is me being crazy again, but I think it's much easier if you have one unbound AST tree that follows the ESTree specs with start and end values, and then again you just track the children and update the existing AST node if needed.

Then you also need to add a bound AST and do some typechecking to validate the ECMAScript syntax.

Something like [this](https://github.com/escaya/escaya/blob/incremental/src/compiler/parser.ts#L75) is what I do now.

TBH! I have no idea atm what I'm doing because I'm still experimenting, but seems to work.

KFlash:
I'm not sure if you can to linting etc with tree-sitter and lexer?

I haven't even pushed half of my source code yet, but my idea is also to do some  linting when I do the type checking, and minification when I do the binding into the finale AST output. The code will also provide some "hints" to help you write better Javascript.

It will all be optional ofc., and if no option set it will do some simple incremental parsing.

PR #83, opened 2020-05-27
fix(scan): token start should skip leading white spaces and comments

### LINKED ISSUE #81: Attach comments
I'm trying to use attachComments functionality of escodegen, which gets tree,comments,tokens as arguments.

Unfortunately, it doesn't preserve the comments using Meriayh and it looks like the tokens types and range are different than what escodegen expect to have.

I tried the following scenario:

`
    
    const sourceCode = `
      //Comment1
      const test1 = 1;
      //Comment2
      const test2 = 2;
    `;
    let tokens = [];
    let comments = [];
    let tree = meriyah.parse(sourceCode, {
      ranges: true,
      globalReturn: true,
      onToken: function (type, start, end) {
        let value = this.tokenValue ? this.tokenValue.toString() : '';
        tokens.push({type: type, value: value, range: [start, end]});
      },
      onComment: function (type, value, start, end) {
        type = type === 'SingleLine' ? 'Line' : type;
        type = type === 'MultiLine' ? 'Block' : type;
        comments.push({type: type, value: value, range: [start, end]});
      }
    });
    replaceStartEndToRange(tree);

    tree = escodegen.attachComments(tree, comments, tokens);
    const res = escodegen.generate(tree, {
      sourceCode: sourceCode,
      format: {
        compact: false,
        preserveBlankLines: true
      },
      comment: true
    });
    expect(res).to.be.equal(sourceCode);
`

sourceCode:
     //Comment1
      const test1 = 1;
      //Comment2
      const test2 = 2;

res: 
     const test1 = 1;

     const test2 = 2;

As you can see, the comments are missing.

I use
"meriyah": "1.9.12"
"escodegen": "1.14.1"

BTW, it works with Esprima.

Thanks,
Niv

3cp:
Fixes #81

### DIFF
--- a/src/lexer/scan.ts
+++ b/src/lexer/scan.ts
@@ -184,7 +184,7 @@ export function nextToken(parser: ParserState, context: Context): void {
   parser.startLine = parser.line;
   parser.token = scanSingleToken(parser, context, LexerState.None);
   if (parser.onToken && parser.token !== Token.EOF)
-    parser.onToken(convertTokenType(parser.token), parser.startPos, parser.index);
+    parser.onToken(convertTokenType(parser.token), parser.tokenPos, parser.index);
 }
 
 export function scanSingleToken(parser: ParserState, context: Context, state: LexerState): Token {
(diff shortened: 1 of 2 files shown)


KFlash:
@3cp Any chance that you could update Meriyah to support the final ESTree decision on the optional chaining? Current implementation is baded on an early draft

3cp:
Sure, I can try.

PR #89, opened 2020-07-18
fix: rename CoalesceExpression

davidbonnet:
The Estree spec extends the LogicalExpression node type by adding
the '??' operator value to it:
https://github.com/estree/estree/blob/master/es2020.md#logicalexpression


KFlash:
Good catch! I guess this is an "left over" from when ESTree discussed this.

Issue #99, opened 2020-09-30
Wrong directive


neumatho:
Hi,

If you parse the expression **"valueOf" in __proto** with the Directives option enabled, then a directive property is added to the ExpressionStatement node, which is wrong. The directive property should only be added for "use strict" literals.

Meriyah also adds a BinaryExpression in the expression property, which is not legal according to the AST specification for directives. See https://github.com/estree/estree/blob/master/es5.md#directive

For these two lines:

```
"use strict";
"valueOf" in __proto
```

the tree looks like:

```
{
  "type": "Program",
  "sourceType": "script",
  "body": [
    {
      "type": "ExpressionStatement",
      "expression": {
        "type": "Literal",
        "value": "use strict",
        "raw": "\"use strict\""
      },
      "directive": "use strict"
    },
    {
      "type": "ExpressionStatement",
      "expression": {
        "type": "BinaryExpression",
        "left": {
          "type": "Literal",
          "value": "valueOf",
          "raw": "\"valueOf\""
        },
        "right": {
          "type": "Identifier",
          "name": "__proto"
        },
        "operator": "in"
      },
      "directive": "valueOf"    <!-- this one should be removed
    }
  ]
}
```

A fix could be that you in the _parseDirective()_ function when checking for the directive option, also test the node type stored in _expression_ variable and only add the directive property if it is a Literal.

-Thomas

KFlash:
cc @3cp @aladdin-add

PR #101, opened 2020-10-06
fix(parser): follow latest estree spec on ExportAllDeclaration

### LINKED ISSUE #97: Named export
Hi,

It seems that Meriyah does not comply with the AST standard when having named export, e.g. **export * as x from 'a'**.

First of all, it creates a _ExportNamedDeclaration_ where it stored an _ExportNamespaceSpecifier_ in the _specifiers_ array. The last one is unknown according to the specifications. If you compare it to Acorn, which seems to do it the right way, then you should create an _ExportAllDeclaration_ and then set the _exported_ property to 'x'.

If you look at the specifications, then you will see that this is the correct way to do it:

https://github.com/estree/estree/blob/master/es2020.md#exportalldeclaration

-Thomas

3cp:
closes #97, #99

-----
Prettier touches few lines of code.

### DIFF
--- a/src/estree.ts
+++ b/src/estree.ts
@@ -68,7 +68,6 @@ export type Node =
   | EmptyStatement
   | ExportAllDeclaration
   | ExportDefaultDeclaration
-  | ExportNamespaceSpecifier
   | ExportNamedDeclaration
   | ExportSpecifier
   | ExpressionStatement
@@ -140,7 +139,6 @@ export type DeclarationStatement =
   | ClassDeclaration
   | ClassExpression
   | ExportDefaultDeclaration
-  | ExportNamespaceSpecifier
   | ExportAllDeclaration
   | ExportNamedDeclaration
   | FunctionDeclaration;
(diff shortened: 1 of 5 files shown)


src/estree.ts
```diff
@@ -415,12 +413,9 @@ export interface EmptyStatement extends _Node {
 export interface ExportAllDeclaration extends _Node {
   type: 'ExportAllDeclaration';
   source: Literal;
+  exported: Identifier | null;
 }
 
-export interface ExportNamespaceSpecifier extends _Node {
-  type: 'ExportNamespaceSpecifier';
```
KFlash, line 421: Has this AST node been removed from the ESTree specs?

3cp: Yes. It is not in estree spec master anymore.

KFlash: I see. What about [AnonymousDefaultExportedFunctionDeclaration](https://github.com/estree/estree/blob/f9d7bd7e6b8d9476f2eaa71d6ca279dff8d49d82/spec.estree#L799) ?

KFlash (approved):

KFlash:
@3cp @aladdin-add can you do a new NPM release?

src/estree.ts
3cp: We seem handled those with normal FunctionDeclaration and ClassDeclaration.

3cp:
I can do the release. Do you mind I add scripts for npm version to generate changelog (we don't have yet), and then publish?

Like this:

```
    "preversion": "npm test",
    "version": "standard-changelog && git add CHANGELOG.md",
    "postversion": "git push && git push --tags && npm publish",
```

KFlash:
No problem if you want to add that 👍

Issue #108, opened 2020-10-13
Wrong range of `TaggedTemplateExpression`


fisker:
```js
require('meriyah').parse('html``', {ranges: true}).body[0].expression
{
  type: 'TaggedTemplateExpression',
  tag: {
    type: 'Identifier',
    name: 'html',
    start: 0,
    end: 4,
    range: [ 0, 4 ]
  },
  quasi: {
    type: 'TemplateLiteral',
    expressions: [],
    quasis: [ [Object] ],
    start: 0,
    end: 6,
    range: [ 0, 6 ]
  },
  start: 4,
  end: 6,
  range: [ 4, 6 ]
}
```

`Node.range` should be `[0, 6]` and `Node.quasi.range` should be `[4, 6]`

PR #114, opened 2020-10-15
fix: fix wrong cooked value in TemplateElement, fix wrong loc and range in various template nodes

### LINKED ISSUE #109: Wrong `TemplateElement` value
```js
require('meriyah').parse('tagA`a`\ntagB`b`', {ranges: true, loc: true}).body[0].expression.quasi.quasis[0].value
{ cooked: 'tagB', raw: 'a' }
```

`Node.value.cooked` should be `a`

3cp:
closes #109, #108

-----

The range/loc in TemplateElement was wrong, because we need to move parser to nextToken before finalise it.
This PR doesn't address null "cooked" value, as indicated in estree spec. I merely added the typing into our estree types.

### DIFF
--- a/src/estree.ts
+++ b/src/estree.ts
@@ -732,7 +732,7 @@ export interface TemplateElement extends _Node {
   type: 'TemplateElement';
   value: {
     raw: string;
-    cooked: string;
+    cooked: string | null;
   };
   tail: boolean;
 }
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -4581,14 +4581,36 @@ export function parseTemplateElement(
   col: number,
   tail: boolean
 ): ESTree.TemplateElement {
-  return finishNode(parser, context, start, line, col, {
+  const node = finishNode(parser, context, start, line, col, {
     type: 'TemplateElement',
     value: {
       cooked,
       raw
     },
     tail
-  });
+  }) as ESTree.TemplateElement;
+
+  const tailSize = tail ? 1 : 2;
+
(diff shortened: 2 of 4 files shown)


KFlash (approved):

3cp:
There is still some difference on range/loc of TemplateLiteral.
For example
```
t`a${b}`
```

The first template literal is `a`, but we use `range:[1,5]`, instead of `range:[2,3]`.

I am not sure how to fix this, because it seems we put the whole `` `a${`` into one token with tokenValue `a`. It might be easier to deal with if separate them into 3 tokens.

KFlash:
Try take a look at the template lexer code, and the rewind code for template when hitting the '{' token.

3cp:
The tokenRaw has following hack
```js
parser.tokenRaw = parser.source.slice(start + 1, parser.index - (token === Token.TemplateSpan ? 1 : 2));
```

I can apply a similar patch to TemplateElement range.
But not sure about patching loc, not sure how newlines are treated around those `` ` ``, `${`, and `}`.

I guess assuming same line?
```
start: {
  line: line,
  column: column + 1,
},
end {
  line: parser.startLine,
  column: parser.startColumn - (parser.token === Token.TemplateSpan ? 1 : 2)
}
```

KFlash:
Sounds reasonable. Give it a try.

3cp:
It looks good.

KFlash:
You merge this and ship a new release?

PR #115, opened 2020-10-16
fix: follow latest decorator proposal

### LINKED ISSUE #105: The class decorators are not correctly positioned in the AST.
The class decorators array is put before the `Class Declaration` element instead of being put as the `decorators` property's value of the `Class Declaration` element.

Here is an example. Given the following JavaScript code:

```javascript
import {MyDecorator} from './decorators.js';
@MyDecorator
class MyViewModel {
}
```

the output AST is:

```javascript
[
  {
    type: 'ImportDeclaration',
    specifiers: [ [Object] ],
    source: {
      type: 'Literal',
      value: './decorators.js',
      start: 26,
      end: 43,
      range: [Array]
    },
    start: 0,
    end: 44,
    range: [ 0, 44 ]
  },
  [
    {
      type: 'Decorator',
      expression: [Object],
      start: 45,
      end: 57,
      range: [Array]
    }
  ],
  {
    type: 'ClassDeclaration',
    id: {
      type: 'Identifier',
      name: 'MyViewModel',
      start: 64,
      end: 75,
      range: [Array]
    },
    superClass: null,
    decorators: [],
    body: { type: 'ClassBody', body: [], start: 76, end: 79, range: [Array] },
    start: 58,
    end: 79,
    range: [ 58, 79 ]
  }
]
```

As you can see the `ClassDeclaration.decorators` property is an empty array. But just before the `ClassDeclaration` object there is an array of decorators.

Removing the lines

```javascript
    case Token.Decorator:
      return parseDecorators(parser, context) as ESTree.Decorator[];
```

from the `parseModuleItem()` function seems to fix the problem.

3cp:
closes #105

-----
Move to new decorator proposal broke the old code `@dec export class A {}`. Need to improve parseExportDeclaration to support decorator before export keyword.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -178,7 +178,12 @@ export function create(
     /**
      * Holds either a function or array used on every token
      */
-    onToken
+    onToken,
+
+    /**
+     * Holds leading decorators before "export" or "class" keywords
+     */
+    leadingDecorators: void 0
   };
 }
 
@@ -415,6 +420,16 @@ export function parseModuleItem(
   line: number,
   column: number
 ): any {
+  if (context & Context.OptionsNext) {
+    const decorators = parseDecorators(parser, context);
+    if (decorators.length) {
+      parser.leadingDecorators = decorators;
+      if (parser.token !== Token.ExportKeyword && parser.token !== Token.ClassKeyword) {
+        report(parser, Errors.InvalidLeadingDecorator);
+      }
+    }
+  }
+
   // ecma262/#prod-ModuleItem
   // ModuleItem :
   //    ImportDeclaration
@@ -2880,6 +2895,11 @@ function parseExportDeclaration(
     // See: https://www.ecma-international.org/ecma-262/9.0/index.html#sec-exports-static-semantics-exportednames
     if (scope) declareUnboundVariable(parser, 'default');
 
+    if (parser.leadingDecorators) {
+      // leadingDecorators should be consumed by parseClassDeclaration
+      report(parser, Errors.InvalidLeadingDecorator);
+    }
+
     return finishNode(parser, context, start, line, column, {
       type: 'ExportDefaultDeclaration',
       declaration
@@ -2892,6 +2912,10 @@ function parseExportDeclaration(
       // 'export' '*' 'as' IdentifierName 'from' ModuleSpecifier ';'
       //
       // See: https://github.com/tc39/ecma262/pull/1174
+
+      if (parser.leadingDecorators) {
+        report(parser, Errors.InvalidLeadingDecorator);
+      }
 
       nextToken(parser, context); // Skips: '*'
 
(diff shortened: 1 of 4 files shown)


3cp:
Added `leadingDecorators` to ParserState, it briefly saves the leading decorators before "export" or "class".

The leading decorators for "class" is only saved in leadingDecorators if it's a module item.
```js
@a @b class C {} // This uses leadingDecorators in ParserState
(@a @b class C {}) // This doesn't use leadingDecorators in ParserState
```

Before module item is finalised, it checks that `leadingDecorators` is consumed by a class declaration, otherwise report an error.

This implementation supports decorators before or after "export". But it also has a side effect, it accepts `@a export @b class C {}` as valid JS, same as `export @a @b class C {}`. I guess legacy decorator spec allows this, but not exactly sure.

src/parser.ts
```diff
@@ -415,6 +420,16 @@ export function parseModuleItem(
   line: number,
   column: number
 ): any {
+  if (context & Context.OptionsNext) {
+    const decorators = parseDecorators(parser, context);
```
KFlash, line 424: If you aim for performance you should inline this into the `parseDecorators` too avoid the length check. Then you could do `let x = 0` before the loop and `x++`; inside the loop and check `if (x !== 0` at the end of the loop. 
Or even faster  ` if(validdecoratorStart) do iteration consume token  `. If no `validdecoratorStart` the loop will not run, so you automatically have an empty array because that's happen outside the `if statement`.

src/parser.ts
```diff
@@ -7906,6 +7932,11 @@ export function parseClassDeclaration(
 
   const decorators: ESTree.Decorator[] = context & Context.OptionsNext ? parseDecorators(parser, context) : [];
 
+  if (parser.leadingDecorators) {
+    decorators.unshift(...parser.leadingDecorators);
```
KFlash, line 7936: Why `unshift`? It's slow

KFlash (approved):

src/parser.ts
3cp: concat?

KFlash: Push?

https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki

src/parser.ts
3cp: Currently leadingDecorators only apply to module item.

By refactoring parseDecorators, do you mean also refactor parseClassElementList to use `parser.leadingDecorators` too? If so, probably I should just rename it to `parser.decorators`.

KFlash: I didn't think about that. My suggestion was just how to get the best performance.

3cp: Just tried. Can not use `parser.decorators` for all decorators because of nested `@bar(@foo class B{}) class A {}`.

The check on length here only applies on the syntax error of a module item. I don't think I could push the logic down to parseDecorators.

3cp:
I did some cleanup.

KFlash (approved):

PR #118, opened 2020-10-24
fix: ForInOfLoopInitializer only applies in strict mode

### LINKED ISSUE #116: for-in cannot be parsed
The following for-in statement cannot be parsed in Meriyah, but works fine in Acorn. The Chrome browser can also parse it when entering it in the console window.

`for(var a = /a/ in []);`

-Thomas

3cp:
closes #116

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2113,10 +2113,7 @@ function parseVariableDeclaration(
       if (
         parser.token === Token.OfKeyword ||
         (parser.token === Token.InKeyword &&
-          (token & Token.IsPatternStart ||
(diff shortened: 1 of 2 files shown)


3cp:
Previously, when webCompact is false (which is default), the ForInOfLoopInitializer error will pop up even in non-strict mode. This fix removed the check on webCompact option.

The error should be only appear in strict mode. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_for-in_initializer

KFlash (approved):

PR #119, opened 2020-10-25
feat: support loc flag for onComment and onToken

### LINKED ISSUE #95: Feature: support `loc` flag when using `onComment`
Setting `{ loc: true, onComment: [] }` should set the `loc` property on generated comment objects.

3cp:
Also fixed onToken, the "start/end" should respect ranges flag, not loc flag.

closes #95

### DIFF
--- a/src/lexer/comments.ts
+++ b/src/lexer/comments.ts
@@ -23,7 +23,15 @@ export function skipHashBang(parser: ParserState): void {
   //   #!  SingleLineCommentChars_opt
   const source = parser.source;
   if (parser.currentChar === Chars.Hash && source.charCodeAt(parser.index + 1) === Chars.Exclamation) {
-    skipSingleLineComment(parser, source, LexerState.None, CommentType.HashBang);
+    skipSingleLineComment(
+      parser,
+      source,
+      LexerState.None,
+      CommentType.HashBang,
+      parser.tokenPos,
+      parser.linePos,
+      parser.colPos
+    );
   }
 }
 
@@ -32,10 +40,13 @@ export function skipSingleHTMLComment(
   source: string,
   state: LexerState,
   context: Context,
-  type: CommentType
+  type: CommentType,
+  start: number,
+  line: number,
+  column: number
 ): LexerState {
   if (context & Context.Module) report(parser, Errors.Unexpected);
-  return skipSingleLineComment(parser, source, state, type);
+  return skipSingleLineComment(parser, source, state, type, start, line, column);
 }
 
 /**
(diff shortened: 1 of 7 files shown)


KFlash (approved):

src/lexer/comments.ts
```diff
@@ -64,18 +66,38 @@ export function skipSingleLineComment(
...
+    endColumn = parser.column;
     end++;
   }
-  if (parser.onComment)
+  if (parser.onComment) {
+    const loc = {
+      start: {
+        // FIXME: there is a bug for HTMLClose.
+        // the start loc of should be before \n-->
+        // which is end of last line.
```
3cp, line 78: Note there is a bug for HTMClose comment, previous range can be easily patch with -4 (the length of `\n-->`), but loc is difficult because the column size of last line is unknown.
The result is the `loc.start` is something like `{line: 2, column: 0}` instead of `{line: 1, column: 14}`.

KFlash: not possible to use a local var to save the value before and after and subtract to get correct value? Why line change? Try local var

3cp: The new line is consumed far before parsing first `-`.

https://github.com/meriyah/meriyah/blob/26d934b50ec018ccf7233ae3974be100178347ad/src/lexer/scan.ts#L351-L362

I am not sure how to save the column info when parsing new line.

https://github.com/meriyah/meriyah/blob/26d934b50ec018ccf7233ae3974be100178347ad/src/lexer/scan.ts#L545-L553

From what I read in ecma262, the starting of HTMClose comment should be before `\n-->`, that's where the line issue came.

3cp: `parser.startLine` and `parser.startColumn` seem recorded last loc before newline.

KFlash: Try remove all instances of `parser.column` and use a `parser.columnIndex` etc. Then do get the `column` value, you can do
`parser.columnIndex - parser.index`. Ofc with `columnAfter` and `columnBefore`

3cp: startLine and startColumn works, but I am running into another trouble.

For multiple comments like
```js
a;
// comment 1
// comment 2
/* comment 3 */
```
They are consumed by a single `scanSingleToken()` call, so the startLIne and startColumn are same for three comments.

I also could not update startLine and startColumn when parsing comments, because that will affect the `loc.end` on previous statement `a;` , because that statement is finalised after next token (all 3 comments) is scanned.

KFlash: what about using startIndex or what the name was? It updated on top of that func

3cp: Wrote some not-pretty code to properly support the loc on HTMLClose comment.

KFlash (approved):

PR #122, opened 2020-10-27
perf(lexer): improved lexer perf

KFlash:


3cp (approved):

KFlash:
@3cp The lexer can be optimized some more, but is it worth it? Still faster than Acorn.

The picture below is from my own private parser code too show off how fast you actually can get a JS parser ;)

experiment

3cp:
That's insane, another difference implementation than escaya?

KFlash:
Yeah the parser used there is my own private code and part of my language. The trick is to get the whitespace and token dispatch code optimized correctly. 
It only does basic parsing and output an compressed AST. The most "heavy" stuff are done in the binder, evaluator and type checking.

KFlash:
Here is an subset of the parser to backup my claim about the benchmark :) All you need to do is `x.parseScript(CODE)`;

https://gist.github.com/KFlash/26926a92e90029622e7c371158963e9e

KFlash:
@3cp The main trick is to limit the while loops and prevent infinite loops. That way the parser will continue to parse / stay alive even if invalid syntax. Throwing an error is expensive :)

See "recovery mode" in Escaya for an idea regarding that. My private parser uses an improved  "recovery mode" with incremental support.

In Meriyah we may be able to optimize some of the loops etc, but I think the token bitwise masks can't be swapped around?

PR #131, opened 2020-10-27
few fixes for jsx

3cp:
These are minor missing code from existing jsx implementation.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -8847,23 +8847,17 @@ export function parseJSXText(
   column: number
 ): ESTree.JSXText {
   scanJSXToken(parser, context);
-  return finishNode(
-    parser,
-    context,
-    start,
-    line,
-    column,
-    context & Context.OptionsRaw
-      ? {
-          type: 'JSXText',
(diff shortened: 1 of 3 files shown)


src/parser.ts
```diff
@@ -8846,11 +8846,24 @@ export function parseJSXText(
...
+      ? {
+          type: 'JSXText',
+          value: parser.tokenValue as string,
+          raw: parser.tokenRaw
+        }
+      : {
+          type: 'JSXText',
+          value: parser.tokenValue as string
+        }
+  );
```
fisker, line 8866:
```suggestion
  const node = {
    type: "JSXText",
    value: parser.tokenValue as string,
  };
  if (context & Context.OptionsRaw) {
    node.raw = parser.tokenRaw;
  }

  return finishNode(parser, context, start, line, column, node);
```

3cp:
@KFlash how is this one?

KFlash (approved):

fisker:
Thanks for the quick fix!

PR #144, opened 2020-10-28
fix: fix infinite loop on broken class body

### LINKED ISSUE #143: Infinite loop in class body
Parsing the following (invalid) class will enter an infinite loop in `parseClassBody`, instead of reporting a syntax error:

```javascript
class Test {
    a, b
}
```
This happens every time you use a comma instead of a semicolon as a field or method delimiter.

3cp:
closes #143

### DIFF
--- a/src/token.ts
+++ b/src/token.ts
@@ -54,7 +54,7 @@ export const enum Token {
   RightBrace   = 15 | IsAutoSemicolon | IsClassField, // }
   RightParen   = 16, // )
   Semicolon    = 17 | IsAutoSemicolon | IsClassField, // ;
-  Comma        = 18 | IsClassField, // ,
+  Comma        = 18, // ,
   LeftBracket  = 19 | IsExpressionStart | IsPatternStart | IsMemberOrCallExpression, // [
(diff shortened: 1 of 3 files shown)


src/token.ts
```diff
@@ -54,7 +54,7 @@ export const enum Token {
   RightBrace   = 15 | IsAutoSemicolon | IsClassField, // }
   RightParen   = 16, // )
   Semicolon    = 17 | IsAutoSemicolon | IsClassField, // ;
-  Comma        = 18 | IsClassField, // ,
+  Comma        = 18, // ,
```
3cp, line 57: @KFlash was this just a fat finger mistake? I want to make sure I didn't miss anything.

KFlash: Dunno. Seems like it. Just remove it. Shouldn't be valid

KFlash:
@3cp Can you swap out current unicode lookup with what you see in this gist? Uses less memory and faster

https://gist.github.com/KFlash/0056e288f7821e1fb985cf4746c15e3a

3cp:
Put it into a todo issue pls.

PR #145, opened 2020-10-28
fix(lexer): \8 \9 are acceptable in web compatibility mode

### LINKED ISSUE #137: Can't parse `"\8"`
```js
const a = "\8";
```

This is valid code.

3cp:
closes #137

### DIFF
--- a/src/lexer/string.ts
+++ b/src/lexer/string.ts
@@ -162,11 +162,6 @@ export function parseEscape(parser: ParserState, context: Context, first: number
       return code;
     }
 
-    // `8`, `9` (invalid escapes)
-    case Chars.Eight:
-    case Chars.Nine:
-      return Escape.EightOrNine;
-
     // HexEscapeSequence
     // \x HexDigit HexDigit
     case Chars.LowerX: {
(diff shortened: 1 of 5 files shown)


KFlash:
Escaped numbers in the grammar is octals ( 0 - 7).  So this shouldn't be allowed. Test it on ASTExplorer with different parsers. Both in sloppy and module goal.

3cp:
No js implementation complain it now. I even checked back on nodejs v0.10.48, it was fine too.
https://github.com/tc39/test262-parser-tests/issues/25

Might be a spec change sometime in the past. The above issue mentioned Annex B.1.2.

3cp:
Almost any invalid escape char is fine in the spec.
http://www.ecma-international.org/ecma-262/11.0/index.html#prod-NonEscapeCharacter

3cp:
I saw many parsers reject it on https://astexplorer.net. But also few accept it, for example typescript.

KFlash:
Then put it under a AnnexB flag? Typescript accept almost everything on ASTExplorer.

3cp:
Isn't AnnexB part of ecma standard. Nodejs accepts it in both strict and non-strict modes.

KFlash:
All parsers supports AnnexB out of the box. Meriyah have an option to turn that off.

3cp:
I didn't see the flag in meriyah code, you might be thinking about another parser of yours.

KFlash:
https://github.com/meriyah/meriyah/blob/master/src/common.ts#L19

KFlash:
Off topic. The `Context` is too long. It already reached `31` and you can risk it turns into negative numbers with 31 and higher.

3cp:
I didn't know webcompat means annex b in meriyah. Will update.

KFlash:
"Additional ECMAScript Features for Web Browsers" = webcompat = AnnexB

KFlash (approved):

PR #157, opened 2020-11-17
Add end position to ParserError

tscpp:


src/common.ts
```diff
@@ -200,34 +200,133 @@ export interface ScopeError {
...
-  end: number;
-  token: Token;
-  onComment: any;
-  onToken: any;
-  tokenValue: any;
-  tokenRaw: string;
+  /**
+   * The source code to be parsed
+   */
+  source: string
```
3cp, line 206: Why remove the semicolon in interface def?

src/errors.ts
```diff
@@ -381,11 +395,22 @@ export class ParseError extends SyntaxError {
  * @param {...string[]} params
  * @returns {never}
  */
-export function report(parser: ParserState, type: Errors, ...params: string[]): never {
-  throw new ParseError(parser.index, parser.line, parser.column, type, ...params);
+export function report(parser: ParserState, context: Context, node: Node, type: Errors, ...params: string[]): never {
+  const end = finishNode(parser, context, parser.index, parser.line, parser.column, node).loc?.end;
```
3cp, line 399: The finishNode mutates nothing on `parser: ParserState`. If you look into the implementation, the `loc.end` is just
```
end: {
        line: parser.startLine,
        column: parser.startColumn
      }
```
Which you can directly get from `parser`.
Also, the availability of `loc.end` in finishNode depends on `OptionsLoc`, but you can always get `parser.startLine` directly.

src/estree.ts
```diff
@@ -5,16 +5,19 @@ export interface _Node {
...
   end: Position;
 }
 
+export interface SourceLocation extends Location {
+  source?: string | null;
+}
+
 export interface Position {
-  /** >= 1 */
+  /** \>= 1 */
```
3cp, line 18: Why add escape to a comment? Probably your editor/tooling?

src/parser.ts
```diff
@@ -191,41 +90,41 @@ export function create(
  * The parser options.
  */
 export interface Options {
-  // Allow module code
+  /** Allow module code */
```
3cp, line 93: Guess this's also your editor/tooling here?

tsconfig.json
```diff
@@ -22,7 +22,8 @@
     "target": "ES2020",
     "resolveJsonModule": true,
     "typeRoots": ["./node_modules/@types"],
-    "types": ["node", "mocha"]
+    "types": ["node", "mocha"],
+    "lib": [ "ES2016" ]
```
3cp, line 26: What was the issue without this config?

src/common.ts
tscpp: Not an actual reason. I'm just used to don't use semis.

src/estree.ts
tscpp: Yes, Visual Studio Code uses `>` as the quote character.

src/parser.ts
tscpp: Using `/** ... */` the comment will be registered by TypeScript as a TSDoc comment. Many editor has support for this. See https://github.com/microsoft/tsdoc

tsconfig.json
tscpp: Using no lib option in compiler options, it will default to `[ "DOM", "ES2016" ]`, and we don't want to have the "DOM" library included because NodeJS has no support for this.

src/parser.ts
3cp: 👍

src/estree.ts
3cp: Do you mean this is needed for TSDoc to generate doc?

src/common.ts
3cp: But they are there for our other interface definitions. I suggest to put it back here for consistency.

I am surprised "prettier" does not have an opinion on this syntax, it accepted both. But there is `semi: true` in our `prettier.config.js`, a "prettier" bug?
If you want to cleanup the type def to remove semicolon, pls create another PR to touch all of the interfaces, if @KFlash has no strong opinion on this syntax thing.

3cp: @fisker should not `semi:true` in `prettier.config.js` enforce the semicolon here?

src/estree.ts
tscpp: Tt's the correct syntax because TSDoc uses markdown.

src/common.ts
fisker: yes, why don't we run `prettier --check` on CI?

tscpp: My bad I didn't run it before the pr.

3cp: @KFlash I am pretty sure there is git hook on pre-commit to check those thing, but I cannot find it in current master head. Did you remove the git hook?

3cp: I tried following locally, prettier didn't pick it up.
```ts
export interface ParserState {
  source: string
  flags: Flags
  index: number;
```

```
huocp@Chunpengs-iMac ~/g/meriyah (master) [1]> npx prettier -c src/
Checking formatting...
All matched files use Prettier code style!
```

KFlash: I havent deleted anything.  The hook should be in the .git folder

3cp: Got it, we got there in `.prettierignore`. @KFlash why we ignored those source files?
```
bench/v1.6.2/
bench/v2/
bench/v8-native-calls.js
node_modules/
dist/
src/token.ts
src/common.ts
src/chars.ts
src/lexer/charClassifier.ts
src/unicode.ts
package-lock.json
```

KFlash: the style get fucked up with Prettier and not lined. Regarding unicode.ts it will be a long list (2000 lines)

KFlash: @3cp Possible to use non-open source to replace Prettier? I got my own pretty printer we could use, but ain't going to open source that one.

3cp: It could be weird for open source project to depend on non-open project.

KFlash: I see your point!

PR #163, opened 2020-12-27
fix(estree): rename FieldDefinition -> PropertyDefinition, PrivateName -> PrivateIdentifier

3cp:
Not sure do we count this as a breaking change for a major release?


src/estree.ts
```diff
@@ -356,21 +356,23 @@ export interface CatchClause extends _Node {
 
 export interface ClassBody extends _Node {
   type: 'ClassBody';
-  body: (ClassElement | FieldDefinition)[];
+  body: (ClassElement | PropertyDefinition)[];
```
KFlash, line 359: This is okay, but I was wondering why the specs are using `PropertyDefinition`. This is part of object literal if you look at the ecma specs. And if you look at the specs for the fields `FieldDefinition` is mentioned there.
The specs mxing two things that isn't related

KFlash (approved):

src/estree.ts
3cp: What do you mean? There is no `FieldDefinition` in current estree spec.

KFlash: That is what I find confusing. It's in the field def .
`PropertyDefinition` has nothing to do with fields. Fields is a Class feature, right?

See the current ECMA specs. It's actually part of ObjectLiteral

https://tc39.es/ecma262/#prod-ObjectLiteral

3cp: That's confusing.

KFlash: Maybe you can drop a comment regarding this in the ESTree repo? My head is elsewhere atm. Working on TS parsing. Maybe something to consider for Meriyah to support?

KFlash:
I guess this is a BC, but do you merge it and publish a new release?

3cp:
Agree. But in urgent need to release.

Issue #168, opened 2021-01-24
ES Next: Logical assignment operators


jods4:
This proposal is completed (stage 4) and supported by TS, it's expected to be part of ES2021.
https://github.com/tc39/proposal-logical-assignment

It also works in most browsers today:
https://caniuse.com/mdn-javascript_operators_logical_nullish_assignment

Can you add support for it?
I have code that uses `??=` and it doesn't parse :(

KFlash:
Try use option `next` for now.

cc / @3cp

jods4:
I use `next: true` in my options.
Many things work, like `import.meta` but `??=` doesn't and I don't see it listed on the README.

Is it supposed to work? I can investigate further if it is, but I have the impression it's not.

KFlash:
What is the code line you try to parse? It's implemented in Meriyah without any need to enable `next`.

You can see it [here](https://github.com/meriyah/meriyah/blob/master/src/parser.ts#L3377-L3384)

jods4:
@KFlash I saw you have a parser of your own. Would you advise using it?
I value ES support, speed and require a standard AST with locations.

KFlash:
I think Meriyah is the best choice.  I have my own parser yes, but it's part of my non-open source project.
It goal was to do the same as Babel with transformers, linting, minifier, prettify, parse TS etc. 

I archived my goal and still playing around.  

You can check out [this readme](https://github.com/kataw/kataw) and try out the demo.

Meriyah is a simple JS parser that follows ESTree and also used in Prettier. It's an better alternative than for example Acorn due to it's performance.

3cp:
Looks like not difficult to implement. Question @KFlash, do we want to shift the numbers below assign operators to make spaces for `||=` `&&=` and `??=`? Or we can give new numbers for them 

```js

  /* Assign operators */
  Assign                  = 29 | IsAssignOp | IsClassField, // =
  ShiftLeftAssign         = 30 | IsAssignOp, // <<=
  ShiftRightAssign        = 31 | IsAssignOp, // >>=
  LogicalShiftRightAssign = 32 | IsAssignOp, // >>>=
  ExponentiateAssign      = 33 | IsAssignOp, // **=
  AddAssign               = 34 | IsAssignOp, // +=
  SubtractAssign          = 35 | IsAssignOp, // -=
  MultiplyAssign          = 36 | IsAssignOp, // *=
  DivideAssign            = 37 | IsAssignOp | IsExpressionStart, // /=
[...]

PR #171, opened 2021-01-29
add pkg.type and pkg.exports

Rich-Harris:
Hi! 

At present, if you import Meriyah as a module in the context of a bundler like Rollup, you have to use named imports otherwise the bundler will complain that there is no default export:

```js
// my-lib/src/index.js
import { parseModule } from 'meriyah';
```

But if you run `my-lib/src/index.js` directly in Node (e.g. because your test suite imports it) then the reverse is true: Node thinks Meriyah is a CommonJS module and thus _only_ has a default export.

This PR adds two things to package.json — a [`"type": "module"` field](https://nodejs.org/api/packages.html#packages_type) and an [`"exports": {...}` map](https://nodejs.org/api/packages.html#packages_exports). The first of these tells Node that the `meriyah` package is ESM rather than CommonJS, which means `import { parseModule }` behaves equivalently whether it's encountered by Node or a bundler. The second is what Node (and bundlers) use to resolve `meriyah` to `/path/to/meriyah/dist/meriyah.esm.js` (or the CommonJS equivalent, if using `require`).

It also renames the `prettier.config.js` and `.eslintrc.js` files to use a `.cjs` extension, since they will otherwise be treated as modules because of this change (and ESLint and Prettier aren't ready for that). It also adds a `package.json` inside the `scripts` directory that causes those files to continue to be treated as CommonJS.

Note that while this will improve compatibility in the future, it is a breaking change so would need a major version bump. Thanks for considering!


3cp:
Interesting, is this the way to support esm in node v12+ without breaking node v10 and before?

Rich-Harris:
(About the failing `npm run lint`: it appears that ESLint is smart enough to detect that the package has `"type": "module"` and configure itself accordingly to error on `require` etc, but not smart enough to detect that there's a `scripts/package.json` that overrides it... I guess the only solution barring an ESLint fix would be to convert everything in `scripts.js` to ESM as well. I can do that if you agree)

Rich-Harris:
> Interesting, is this the way to support esm in node v12+ without breaking node v10 and before?

Exactly, yes. For anyone doing this...

```
const meriyah = require('meriyah');
```

...nothing will change, whether they're using Node 10 or Node 15. This change only affects people who are using Node 12 or above and want to use `import` syntax without a build step.

KFlash (approved):

3cp:
Updating scripts to esm means requiring nodejs12+ to build. Maybe we just skip scripts folder for eslint? @KFlash ?

@Rich-Harris thanks for showing this neat trick. I was quite upset about nodejs esm support because it is not backward compatible.

3cp:
Oh, it's using .cjs on eslint on prettier, already requiring nodejs v12+ to build.

3cp:
I will skip the folder for eslint.

Rich-Harris:
thanks so much for the quick response!

3cp:
Will take few days before next release. Currently adding #168.

3cp:
@Rich-Harris I got trouble to consume meriyah in commonjs code.

In a folder, I do
```
npm i ../../github/meriyah
node test.cjs
```

The test.cjs is just
```js
const {parseScript} = require('meriyah');
console.log(parseScript('a=1'));
```

Nodejs v14.15.4 complains
```
huocp@chunpengs-mbp ~/p/t> node test.cjs
internal/modules/cjs/loader.js:1080
      throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
      ^

Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/huocp/github/meriyah/dist/meriyah.umd.js
require() of ES modules is not supported.
require() of /Users/huocp/github/meriyah/dist/meriyah.umd.js from /Users/huocp/playground/t/test.cjs is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename meriyah.umd.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/huocp/github/meriyah/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/Users/huocp/playground/t/test.cjs:1:23)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14) {
  code: 'ERR_REQUIRE_ESM'
}
```

It seems thinking the `exports.require: './dist/meriyah.umd.js'` is an ESM module??

3cp:
Ok, `exports.require: './dist/meriyah.umd.cjs'` and copy `meriyah.umd.js to meriyah.umd.cjs` fixed the issue.

Rich-Harris:
Ah, of course — apologies for missing that, good catch!

3cp:
v4.1.1 should work as expected.

PR #172, opened 2021-01-29
chore: ignore scripts/ for eslint for now

3cp:


KFlash (approved):

KFlash:
It will still be possible to "lint" Meriyah after this? I'm not using ESLint anymore privately so lots of things changed since I used it so not 100% sure about it settings

PR #173, opened 2021-01-30
feat: add support of logical assignment ||=, &&=, and ??=

### LINKED ISSUE #168: ES Next: Logical assignment operators

3cp:
closes #168


src/parser.ts
```diff
@@ -7984,7 +7984,7 @@ export function parseClassExpression(
 
   nextToken(parser, context);
 
-  if (((parser.token & 0x10ff) ^ 0x54) > 0x1000) {
+  if (((parser.token & 0x10ff) ^ 0x57) > Token.Keyword) {
```
3cp, line 7987: This code should be refactored using Token constants from hard coded binary.
I don't fully understand what's doing right now, but since we shift 3 numbers for many tokens, this line doesn't work right now. And my attempt of fix (by adding 3 to 0x54) didn't work. I assume the 0x54 (84) means lower part of ExtendsKeyword (now is 87).

We need some comments to explain the code, plus a proper fix. @KFlash

3cp: I think I understand why it fails now.
The previous value for ExtendsKeyword 84 is an even number, so the lowest bit is 0, the Token.Identifier is 1, `1 ^ 0` will work.
But now ExtendsKeyword 87 has 1 at lowest bit. It doesn't work with Token.Identifier `1 ^ 1 === 0`.
That also tells me the previous code is too tricky.

3cp: Is the logic " if this token is a keyword and not the `extends` "?

3cp: OK, fixed. There are two places for the same code, one in parseClassDeclaration, one in parseClassExpression.

KFlash: Your way of thinking is correct. How did you fix it?

KFlash: > This code should be refactored using Token constants from hard coded binary.

How would you do that? Use of constants?

3cp: Pls review the latest code.

src/parser.ts
```diff
@@ -7876,10 +7876,7 @@ export function parseClassDeclaration(
 
   const { tokenValue } = parser;
 
-  if (
-    ((parser.token & 0b0000000000000000001_0000_11111111) ^ 0b0000000000000000000_0000_01010100) >
-    0b0000000000000000001_0000_00000000
-  ) {
+  if (parser.token & Token.Keyword && parser.token !== Token.ExtendsKeyword) {
```
3cp, line 7879: Please view these two lines.

src/parser.ts
```diff
@@ -7984,7 +7981,7 @@ export function parseClassExpression(
 
   nextToken(parser, context);
 
-  if (((parser.token & 0x10ff) ^ 0x54) > 0x1000) {
+  if (parser.token & Token.Keyword && parser.token !== Token.ExtendsKeyword) {
```
3cp, line 7984: Please view these two lines.

src/parser.ts
KFlash: Sure this will work?  Why not  

`if (parser.token & (Token.IsIdentifier | Token.FutureReserved) && parser.token !== Token.ExtendsKeyword ) {`

Keyword will make this case valid `class switch() {}`.

The two token can be merged into signed integer (32-bit), but maybe better to keep it as it is for readability?

src/parser.ts
KFlash: See my  prev comment for same case.

src/parser.ts
3cp: I didn't quite understand the existing line, it is confusing for me. The existing unit tests didn't complain about my code. I will add your test case and try again.

3cp: The throw is inside the if branch, my code still invalids `class switch() {}`, `class eval {}` `class arguments {}`.
```js

  if (parser.token & Token.Keyword && parser.token !== Token.ExtendsKeyword) {
    if (isStrictReservedWord(parser, context, parser.token)) {
      report(parser, Errors.UnexpectedStrictReserved);
    }

    if ((parser.token & Token.IsEvalOrArguments) === Token.IsEvalOrArguments) {
      report(parser, Errors.StrictEvalArguments);
    }
```

```
huocp@chunpengs-mbp ~/g/meriyah (logic-assign)> npx ts-node
> import {parseModule} from './src/meriyah'
{}
> var r = parseModule('class switch() {}')
[/Users/huocp/github/meriyah/src/errors.ts:187
    throw new ParseError(parser.index, parser.line, parser.column, type, ...params);
    ^

Uncaught SyntaxError: [1:12]: Unexpected strict mode reserved word
] {
  index: 12,
  line: 1,
  column: 12,
  description: '[1:12]: Unexpected strict mode reserved word',
  loc: { line: 1, column: 12 }
}
> 
```

Added test.

3cp: This confused me: identifier is also considered a keyword?

```ts
  Keyword                 = 1 << 12,
  Contextual              = 1 << 13 | Keyword,

  IsIdentifier             = 1 << 17 | Contextual,
```

3cp: @KFlash can you review the 2 posts above?

KFlash: That it throws on invalid case is correct.  The bitwise masks seems weird.  Contextual should be future reserved I guess or plain ident.

3cp: merge this code?

KFlash: you can

KFlash (approved):

PR #174, opened 2021-01-30
docs: update readme and package.json

KFlash:


3cp (approved):

3cp:
ci said package-lock is not in sync, pls do `npm i` and commit lock file.

package.json
```diff
@@ -13,7 +13,6 @@
   },
   "types": "dist/src/meriyah.d.ts",
   "typings": "dist/src/meriyah.d.ts",
-  "author": "KFlash",
```
3cp, line 16: Removed author but didn't add to authors?

3cp:
Maybe just add package-lock to gitignore?

package.json
KFlash: > ci said package-lock is not in sync, pls do `npm i` and commit lock file.

I can't do it from a cp. Can you fix it?

> Removed author but didn't add to authors?

Didn't figure out how to write it in authors. It's only a nick that one. Can you take a stab at it?

KFlash:
> Maybe just add package-lock to gitignore?

Is it safe?

3cp:
The worst result is just some 3rd party dep breaking our CI, hardly any effect on end users.

Removing lock file may help us to detect 3rd party breaking change earlier.

Lock file makes more sense in app project which you want to build and deploy with confidence.

KFlash:
You can fix it? BTW. Circleci got a problem?

3cp:
package.json has author and contributors, no authors. https://docs.npmjs.com/cli/v6/configuring-npm/package-json#people-fields-author-contributors

KFlash:
I think this can be merged ?

PR #175, opened 2021-01-31
fix(jsx): decode html entities for JSXText value

### LINKED ISSUE #133: TODO process JSXText value
Copied from #129
```html
<div>&nbsp;</div>
```
Should be `{value: " ", raw: "&nbsp;"}`, but we didn't process value `{value: "&nbsp;", raw: "&nbsp;"}`.

Reference: https://github.com/fb55/entities

3cp:
closes #133


src/lexer/decodeHTML.ts
```diff
@@ -0,0 +1,2185 @@
...
+  });
+}
+
+function decodeCodePoint(codePoint: number): string {
+  if ((codePoint >= 0xd800 && codePoint <= 0xdfff) || codePoint > 0x10ffff) {
+    return '\uFFFD';
+  }
+
+  if (codePoint in decodeMap) {
+    codePoint = decodeMap[codePoint];
```
3cp, line 2181: I have not found the spec for this special codePoint mapping. Basically copied form fb55/entities implementation now.

KFlash: eh wow. That code consist of 2k lines? Can you check how Acorn JSX plugin does this?

3cp: acron-jsx seems only uses part of the entities map https://github.com/acornjs/acorn-jsx/blob/master/xhtml.js

I checked few of their missed ones, React app still support them, obviously acron-jsx deliberated skipped non-popular entities.

3cp: acron-jsx also didn't patch number code point like fb55/entities (This line I noted in this thread).
The decodeMap is a small object to translate few special code points, I have not found any spec for it.

I tested in React app, the decodeMap patch is necessary to match real app behaviour.

KFlash (approved):

src/lexer/decodeHTML.ts
```diff
@@ -0,0 +1,2185 @@
...
+
+// Used code from https://github.com/fb55/entities (BSD license)
+export function decodeHTMLStrict(text: string): string {
+  return text.replace(/&(?:[a-zA-Z]+|#[xX][\da-fA-F]+|#\d+);/g, (key) => {
+    if (key.charAt(1) === '#') {
+      const secondChar = key.charAt(2);
+      if (secondChar === 'X' || secondChar === 'x') {
+        return decodeCodePoint(parseInt(key.substr(3), 16));
+      }
+      return decodeCodePoint(parseInt(key.substr(2), 10));
```
fisker, line 2169:
```suggestion
      let codePoint;
      if (secondChar === 'X' || secondChar === 'x') {
        codePoint = parseInt(key.slice(3), 16);
      } else {
        codePoint = parseInt(key.slice(2), 10);
      }
      return decodeCodePoint(codePoint);
```

fisker, line 2169:
```suggestion
    const codePoint =
      secondChar === 'X' || secondChar === 'x'
        ? parseInt(key.slice(3), 16)
        : parseInt(key.slice(2), 10);
    return decodeCodePoint(codePoint);
```

src/lexer/jsx.ts
```diff
@@ -96,8 +97,9 @@ export function scanJSXToken(parser: ParserState, context: Context): Token {
         if (CharTypes[parser.currentChar] & CharFlags.JSXToken) break;
       }
 
-      parser.tokenValue = parser.source.slice(parser.tokenPos, parser.index);
-      if (context & Context.OptionsRaw) parser.tokenRaw = parser.tokenValue;
+      const raw = parser.source.slice(parser.tokenPos, parser.index);
+      if (context & Context.OptionsRaw) parser.tokenRaw = raw;
+      parser.tokenValue = decodeHTMLStrict(raw);
```
fisker, line 102: I wonder how this impact performance, maybe use a `getter`?

KFlash: easier to hoist `const { tokenValue, tokenPos, fisker, tokenRaw } = parser` and re-use

3cp: Not sure, changing to getter might touch lots of code. @KFlash?

src/lexer/decodeHTML.ts
```diff
@@ -2163,10 +2163,11 @@ export function decodeHTMLStrict(text: string): string {
   return text.replace(/&(?:[a-zA-Z]+|#[xX][\da-fA-F]+|#\d+);/g, (key) => {
     if (key.charAt(1) === '#') {
       const secondChar = key.charAt(2);
-      if (secondChar === 'X' || secondChar === 'x') {
-        return decodeCodePoint(parseInt(key.substr(3), 16));
-      }
-      return decodeCodePoint(parseInt(key.substr(2), 10));
+    const codePoint =
+      secondChar === 'X' || secondChar === 'x'
```
KFlash, line 2167: what is 'x' ?  Note that this can be simplified and force to lowercase `secondChar | 32 === codePoint`. If no codepoint yo u can do the same trick with 0x20

KFlash (approved):

Issue #141, opened 2020-10-28
`main` field in `package.json`


fisker:
The `browser` field already pointed to `meriyah.umd.js`, shouldn't `main` point to `meriyah.cjs.js`?

And `browser` might want use the minified version?

3cp:
I guess the benefit is that the umd main allows jsdelivr (or unpack) to serve umd as default main. So the short url can be used in `<script>` tag without any module loader.

https://cdn.jsdelivr.net/npm/meriyah@3.1.0 is same as the full path https://cdn.jsdelivr.net/npm/meriyah@3.1.0/dist/meriyah.umd.min.js

fisker:
As I understand `jsdelivr` looks for `browser`

https://github.com/prettier/prettier/pull/8387#issuecomment-633232840

We can add `unpkg` field for `unpkg`

fisker:
Configuring a default file in package.json
For packages hosted on npm, we support setting a "default" file for each package. This file is displayed at the top of directory listings on our website and available under shorter URL on the CDN. The default file can be configured by setting one of the following fields in package.json (ordered by priority):

jsdelivr
browser
main

Copied from https://www.jsdelivr.com/features

3cp:
Thx! Good to know. It's @KFlash's call.
Note this is a breaking change.

benmccann:
I am unable to use Meriyah 4.1.1 in my project. It looks like this may have been attempted to be fixed in https://github.com/meriyah/meriyah/pull/171, but it doesn't seem sufficient for me.

```
node_modules/.pnpm/meriyah@4.1.1/node_modules/meriyah/dist/meriyah.umd.js:1
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: node_modules/.pnpm/meriyah@4.1.1/node_modules/meriyah/dist/meriyah.umd.js
require() of ES modules is not supported.
require() of node_modules/.pnpm/meriyah@4.1.1/node_modules/meriyah/dist/meriyah.umd.js from packages/kit/src/api/dev/transform.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename meriyah.umd.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from node_modules/.pnpm/meriyah@4.1.1/node_modules/meriyah/package.json.
```

[...]

PR #177, opened 2021-02-09
Fix CommonJS entry point in package.json

benmccann:
Closes https://github.com/meriyah/meriyah/issues/141

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "type": "module",
   "exports": {
     "import": "./dist/meriyah.esm.js",
-    "require": "./dist/meriyah.umd.cjs"
+    "require": "./dist/meriyah.cjs"
   },
   "types": "dist/src/meriyah.d.ts",
   "typings": "dist/src/meriyah.d.ts",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "meriyah",
   "version": "4.1.1",
   "description": "A 100% compliant, self-hosted javascript parser with high focus on both performance and stability",
-  "main": "dist/meriyah.umd.js",
+  "main": "dist/meriyah.umd.cjs",
   "module": "dist/meriyah.esm.js",
   "jsnext:main": "dist/meriyah.esm.js",
   "browser": "dist/meriyah.umd.js",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "meriyah",
   "version": "4.1.1",
   "description": "A 100% compliant, self-hosted javascript parser with high focus on both performance and stability",
-  "main": "dist/meriyah.cjs.js",
+  "main": "dist/meriyah.cjs",
   "module": "dist/meriyah.esm.js",
   "jsnext:main": "dist/meriyah.esm.js",
   "browser": "dist/meriyah.umd.js",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "meriyah",
   "version": "4.1.1",
   "description": "A 100% compliant, self-hosted javascript parser with high focus on both performance and stability",
-  "main": "dist/meriyah.umd.cjs",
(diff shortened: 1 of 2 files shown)


KFlash (approved):

fisker:
What about this https://github.com/meriyah/meriyah/issues/141#issuecomment-717716317?

benmccann:
> What about this #141 (comment)?

sure. I've added an `unpkg` field as well now

package.json
```diff
@@ -2,10 +2,11 @@
   "name": "meriyah",
   "version": "4.1.1",
   "description": "A 100% compliant, self-hosted javascript parser with high focus on both performance and stability",
-  "main": "dist/meriyah.umd.js",
+  "main": "dist/meriyah.umd.cjs",
   "module": "dist/meriyah.esm.js",
   "jsnext:main": "dist/meriyah.esm.js",
   "browser": "dist/meriyah.umd.js",
+  "unpkg": "dist/meriyah.umd.js",
   "type": "module",
```
aladdin-add, line 10: should remove `"type": "module"` - as it also has cjs?

benmccann: No, `"type": "module"` just means that `.js` files will be loaded as ES modules. It doesn't prevent you from also shipping `cjs`. In fact the `.cjs` ending is more useful when you have `"type": "module"` because then you're specifying that specific file should be loaded as CommonJS and not as the ES default specified by `"type": "module"`

aladdin-add: I see.  just feel a little counter-intuitive for both shipping cjs and esm.

consider renaming `.esm.js` => `.mjs` , just like `.cjs`?

benmccann: It's pretty common to ship both multiple versions so that the appropriate one is available for each consumer. I think it makes sense how it is

aladdin-add: to clarify: it is  common to ship both multiple versions, but a little counter-intuitive to have `"type": "module"` in this kind of packages. 😄

aladdin-add: well. I think I was sold!

aladdin-add (approved):

3cp:
Isn't #141 is about cjs as main, not umd?

package.json
```diff
@@ -2,10 +2,11 @@
   "name": "meriyah",
   "version": "4.1.1",
   "description": "A 100% compliant, self-hosted javascript parser with high focus on both performance and stability",
-  "main": "dist/meriyah.umd.js",
+  "main": "dist/meriyah.umd.cjs",
```
3cp, line 5: This change is not necessary. For nodejs supports `exports.import/require`, it will ignore main field.
For old nodejs doesn't support `exports.import/require`, this main field works fine.

I tested current release with nodejs v10 (doesn't support `exports.import/require`), v12, v14, all worked fine.

3cp: I think you mistook this file as cjs, the real cjs file is `meriyah.cjs.js`.

The `.cjs` in `*.umd.cjs` is just for making nodejs happy because this package is marked as `"type": "module"`, nodejs (v12+) thinks `meriyah.umd.js` should be an esm file, so we added `.cjs` to calm down nodejs.

3cp: You can update both "main" and "exports.require" to `meriyah.cjs`, also need to update rollup setup to change the file name from `meriyah.cjs.js` to `meriyah.cjs`.

benmccann: done

package.json
```diff
@@ -2,14 +2,15 @@
   "name": "meriyah",
   "version": "4.1.1",
   "description": "A 100% compliant, self-hosted javascript parser with high focus on both performance and stability",
-  "main": "dist/meriyah.umd.js",
+  "main": "dist/meriyah.cjs.js",
```
3cp, line 5: Based on discussion on #141, let's use meriyah.cjs here.

benmccann: done

3cp:
v4.1.2 is out.

PR #184, opened 2021-03-04
fix(parser): fix missing rejection on function name

### LINKED ISSUE #182: Meriyah does not properly follow ECMA standards on function declaration and calling.
When experimenting with Meriyah, I discovered the parser properly registers the following invalid function declarations and calls:

```
function 10() {
   return 10;
}

10();
```

```
function "string"() {
   return "string"();
}

"string"();
```

Meriyah also struggles when calling functions. It will assume that the name of the function being called is an identifier, and not a literal.

3cp:
closes #182

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -4817,7 +4817,11 @@ export function parseFunctionDeclaration(
 
     firstRestricted = parser.token;
 
-    id = parseIdentifier(parser, context, 0);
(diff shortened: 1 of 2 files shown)


3cp:
Not finished. I need to reject function call too.

3cp:
@KFlash ready for review.

KFlash (approved):

PR #188, opened 2021-04-08
fix(parser): rejects "await 2**2"

### LINKED ISSUE #187: Should throw on `async () => await 5 ** 6;`
Babel PR https://github.com/babel/babel/pull/12441

3cp:
closes #187

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3613,6 +3613,8 @@ export function parseAwaitExpression(
       parser.colPos
     );
 
+    if (parser.token === Token.Exponentiate) report(parser, Errors.InvalidExponentiationLHS);
+
     parser.assignable = AssignmentKind.CannotAssign;
 
     return finishNode(parser, context, start, line, column, {
--- a/test/parser/expressions/exponentiation.ts
+++ b/test/parser/expressions/exponentiation.ts
@@ -80,7 +80,8 @@ describe('Expressions - Exponentiation', () => {
     ['(!3 ** 2)', Context.None],
     ['(+x ** 2)', Context.None],
     ['(a * +a ** a ** 3)', Context.None],
-    ['for (var import.meta of [1]) {}', Context.None]
+    ['for (var import.meta of [1]) {}', Context.None],
+    ['async function f() { await 2 ** 2; }', Context.None]
   ]);
 
   for (const arg of [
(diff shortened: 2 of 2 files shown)


3cp:
While this seems like a simple fix in meriyah code, but I noticed our parsePrimaryExpression->parseUnaryExpression didn't handle AwaitExpression (it's handled in parsePrimaryExpression->parseAwaitExpression).

https://github.com/meriyah/meriyah/blob/da41eba70aae5421c32cc692cc2f07b5d1c7adca/src/parser.ts#L4164-L4167

https://github.com/meriyah/meriyah/blob/da41eba70aae5421c32cc692cc2f07b5d1c7adca/src/parser.ts#L4211-L4218

@KFlash I am not sure do we need to move the code path parsePrimaryExpression->parseAwaitExpression to be covered by parsePrimaryExpression->parseUnaryExpression as AwaitExpression should be an UnaryExpression?
That seems to be a much bigger change (but closer to spec).

3cp:
Ecma PrimaryExpression can be AwaitExpression, I am not clear about whether AwaitExpression should be covered by parseUnaryExpression.

KFlash:
I will investigate and come back to you

KFlash:
This PR seems okay. I think you can use 'parseLeftHandSideExpression' in Await expr instead of parseExpression to narrow things down.

KFlash:
@3cp Plan to merge this one?

3cp:
It's waiting for your review approval:-)

KFlash (approved):

KFlash:
LGTM. I'm using an similar solution in Kataw.

KFlash:
@3cp 你怎么看“Kataw”？ 你对如何改进它有什么想法吗？

3cp:
我没有深入去看kataw，因为我还只能用estree ast。换ast要重写太多代码。

KFlash:
你在用 AST 做什么？Kataw 正在使用 CST。Pretty printing 即将推出。 ESTree很慢

3cp:
I know estree is not fast. I am using it in a bundler dumberjs/dumber using AMD module format.
I like AMD because it supports dynamic runtime features naturally. Webpack and other alternatives don't have a real module loader at runtime, so they have very unnatural implementation of dynamic behavior such as code split.

ESM format was designed to be statically analysible. But now they are adding top level await in order to support dynamic module importing. It's just complicating the whole. If they design ESM as an asynchronous module loader from beginning, there would not be so much work now.

AMD got that right 20 yrs ago, it's tiny and work as expected. Now we are reimplementing it but with a much bloated design and implementation.

KFlash:
I see. That's on my TODO for the future. A bundler with various formats.

PR #197, opened 2021-07-11
feat(parser): support top-level await

3cp:


3cp:
Running into an issue. The meriyah `Context.InGlobal` is not same as kataw's `Context.TopLevel`.
Notably the parseParenthesizedExpression clears Context.InGlobal on context.

So right now, it still rejects top-level await if it's in a parenthesized expression. `const foo = (await bar);`

KFlash:
Why that bitwise mask got unset? Try set it back and see if any tests breaks.

KFlash:
Why that bitwise mask got unset? Try set it back and see if any tests breaks.

3cp:
I thought the same.
After I changed this line in parseParenthesizedExpression
```js
  // context = (context | Context.DisallowIn | Context.InGlobal) ^ (Context.InGlobal | Context.DisallowIn);
  context = (context | Context.DisallowIn) ^ Context.DisallowIn;
```

Strangely it now failed to reject this code `['({ def\\u0061ult })', Context.None],`, it should trigger an invalid escaped keyword "default".

KFlash:
I think that is caused by the fact you have used 31 "slots" for context masks and it turns into negative value with side effects

3cp:
I thought `OptionsUniqueKeyInPattern = 1 << 31,` needs to be explicitly turned on to have that issue?

KFlash:
it depends on TS. Rare issues happens with TS all the time when it comes to enum values. Anyway. Non-failing escape keyword test is ok vs current issue

3cp:
I agree, I will just block that test.

src/errors.ts
```diff
@@ -274,7 +274,6 @@ export const errorMessages: {
   [Errors.IllegalReturn]: 'Illegal return statement',
   [Errors.InvalidForLHSBinding]: 'The left hand side of the for-header binding declaration is not destructible',
   [Errors.InvalidNewTarget]: 'new.target only allowed within functions',
-  [Errors.InvalidEscapedKeyword]: "'Unexpected token: 'escaped keyword'",
```
3cp, line 277: This is a duplicated entry of line 207: `[Errors.InvalidEscapedKeyword]: "Unexpected token: 'escaped keyword'"`

3cp:
Somehow the circleci tests are not showing in this PR checks.

KFlash:
isn't that a circleci issue?

3cp:
No idea. I don't use circleci personally.

KFlash:
just merge this as is ?

3cp:
Sure, if you found the code change is reasonable.

KFlash (approved):

KFlash (approved):

PR #201, opened 2022-02-10
fix(parser): precedence of `<=` and `>=` operators

DimaIT:
Fixes invalid precedence of `<=` and `>=` operators

According to [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table) `<=` and `>=` operators have precedence same as `<` and `>` and greater than equal operators.

At the moment expression `a >= b !== c >= d` parsed as `((a >= b) !== c) >= d`, but expected `(a >= b) !== (c >= d)`.


3cp (approved):
Thx!
LGMT @KFlash

3cp:
@KFlash circleci seems not running. I can convert to GitHub Actions if you are fine with it.

PR #206, opened 2022-03-02
Allow HTML comments in modules contexts

e9x:
This pull request will add the `htmlCommentModule` option that will allow HTML comments (`<!-- COMMENT -->`) when the module option is enabled. Although browsers forbid modules from having HTML comments in 
module sources, this will allow for more lenient parsing in cases where the source type being a module is unknown.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -39,6 +39,7 @@ export const enum Context {
   OptionsSpecDeviation = 1 << 29,
   AllowEscapedKeyword = 1 << 30,
   OptionsUniqueKeyInPattern = 1 << 31,
+  OptionsHTMLCommentModule = 1 << 32,
 }
 
 /**
--- a/README.md
+++ b/README.md
@@ -79,6 +79,9 @@ This is the available options:
   // The flag to allow return in the global scope
   globalReturn: false;
 
+  // Allow HTML comments when the module option is `true`
+  htmlCommentModule: false;
+
   // The flag to enable implied strict mode
   impliedStrict: false;
 
--- a/src/lexer/comments.ts
+++ b/src/lexer/comments.ts
@@ -48,7 +48,7 @@ export function skipSingleHTMLComment(
   line: number,
   column: number
 ): LexerState {
-  if (context & Context.Module) report(parser, Errors.Unexpected);
+  if (context & Context.Module && !(context & Context.OptionsHTMLCommentModule)) report(parser, Errors.Unexpected);
   return skipSingleLineComment(parser, source, state, type, start, line, column);
 }
 
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -227,6 +227,8 @@ export interface Options {
   onToken?: OnToken;
   // Creates unique key for in ObjectPattern when key value are same
   uniqueKeyInPattern?: boolean;
+  // Allow HTML comments when the module option is `true`
+  htmlCommentModule?: boolean;
 }
 
 /**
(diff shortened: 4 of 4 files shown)


fisker:
Isn't this enabled behind `webcompat`?

src/common.ts
```diff
@@ -39,6 +39,7 @@ export const enum Context {
   OptionsSpecDeviation = 1 << 29,
   AllowEscapedKeyword = 1 << 30,
   OptionsUniqueKeyInPattern = 1 << 31,
+  OptionsHTMLCommentModule = 1 << 32,
```
3cp, line 42: If you check the value of `1 << 32` in node, it returns `1`. This is due to JS uses 32 bits integer for `<<` operator (store float 1 in 32 bits integer, do shift, then convert back to float).
So `1 << 32` is same as OptionNext `1 << 0`.
We have exhausted the 32 bits for all options. I am not sure how to add one more option, if we give up binary mask, we will loose performance for option check.
As @fisker suggested, putting it behind webcompat might be an option.

3cp: What other parsers do?

3cp (changes requested):

3cp:
> allow for more lenient parsing in cases where the source type being a module is unknown

Why use `parseModule` when the type is unknown?

src/common.ts
fisker: > What other parsers do?

Babel throws when `sourceType: module`,  only parse when `sourceType` set to `script` or `unambiguous`.

e9x: > We have exhausted the 32 bits for all options. I am not sure how to add one more option, if we give up binary mask, we will lose performance for option check.

I feel like losing performance is an acceptable trade-off for more options. Perhaps the context could be an object, providing similar-if not better performance.

e9x:
> > allow for more lenient parsing in cases where the source type being a module is unknown
> 
> Why use `module: true` when the type is unknown?

I specify the source as a module to allow the script to contain import/export declarations. Although I could instead determine if every script is a module, I feel like making HTML comments allowed is a more beneficial solution.

Putting this option behind webcompat will create confusion. I see webcompat allows for HTML comments in the source. The default behavior in browsers when a module script contains comments is to throw an error, which is what meriyah does by default.

Perhaps I am just lazy and should accommodate for module scripts in my code. Determining if a script is a module or normal script will add more complexity than its worth.

src/common.ts
e9x: If this PR isn't adding bulk, I will attempt to make the context an object. I believe this will have benefits beyond adding more options. Doing a binary expression when checking an option may be less efficient when compared to reading a property.

3cp:
As @fisker noted, other parsers also throw when parsing in module mode. I am not convinced allowing HTML comment in module mode is a right decision. It sounds like a deviation of spec.

e9x:
> As @fisker noted, other parsers also throw when parsing in module mode. I am not convinced allowing HTML comment in module mode is a right decision. It sounds like a deviation from spec.

I agree, I may be approaching this the wrong way.

I noticed Acorn has an option named `allowImportExportEverywhere`. This option would accomplish the same thing, import/export in sources, while allowing HTML comments (`module` isn't required to be true). I believe this could have more use outside of my scenario. Could Meriyah can have an option similar to this?

3cp:
I think you can just use existing option:
```
// Allow edge cases that deviate from the spec
specDeviation: false
```

Sounds like designed for this kind of thing :-)

3cp:
I meant you can use specDeriation to implement this, to allow html comments in module mode.
Without adding new option.

e9x:
> I meant you can use specDeriation to implement this, to allow html comments in module mode. Without adding new option.

Ah. I will reopen this PR when I can implement this.

3cp:
BTW, pls add 2 test cases too.
module mode, throws.
module mode + spec deviation, parses.

PR #210, opened 2022-03-18
fix(parser): super call should be allowed in private method

### LINKED ISSUE #203: Unexpected private field
I get an error of `Unexpected private field` when using the `super` keyword in a class function with a private name. It works fine if the method is not private, i.e. no hash (`#`) in front of the method name. Here's a minimal example with a comparison to `acorn`.
```
import { parse } from 'meriyah';
import { parse as aparse } from 'acorn';

const test = `class A extends B
{
    #xyz()
    {
        super.abc();
    }
}
`;

console.log( `------------------------ ACORN` );
console.log( aparse( test, { ecmaVersion: 2022 }) );
console.log( `------------------------ MERIYAH` );
console.log( parse( test, { next: true }) );
```
which produces the following output:
```
------------------------ ACORN
Node {
  type: 'Program',
  start: 0,
  end: 66,
  body: [
    Node {
      type: 'ClassDeclaration',
      start: 0,
      end: 65,
      id: [Node],
      superClass: [Node],
      body: [Node]
    }
  ],
  sourceType: 'script'
}
------------------------ MERIYAH
file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:181
    throw new ParseError(parser.index, parser.line, parser.column, type, ...params);
          ^

ParseError [SyntaxError]: [5:14]: Unexpected private field
    at report (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:181:11)
    at parseSuperExpression (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:6090:17)
    at parsePrimaryExpression (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:6355:20)
    at parseExpressionOrLabelledStatement (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:4935:20)
    at parseStatement (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:4920:20)
    at parseStatementListItem (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:4870:20)
    at parseFunctionBody (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:6061:19)
    at parseMethodDefinition (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:6994:18)
    at parseClassElementList (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:8407:19)
    at parseClassBody (file:///home/.../project/node_modules/meriyah/dist/meriyah.esm.js:8285:19) {
  index: 51,
  line: 5,
  column: 14,
  description: '[5:14]: Unexpected private field',
  loc: { line: 5, column: 14 }
}
```

3cp:
Also fixed the error message for invalid super call.

closes #203

### DIFF
--- a/test/parser/next/private_methods.ts
+++ b/test/parser/next/private_methods.ts
@@ -192,7 +192,9 @@ describe('Next - Private methods', () => {
     'foo() { this.#m, (() => this)().#m }',
     'foo() { this.#m, (() => this)().#m }',
     'foo() { this.#m, (() => this)().#m }',
-    'method() { super.#x(); }'
+    'method() { super.#x(); }',
+    '#method() { super.x(); }',
+    '#method() { super.#x(); }'
   ]) {
     it(`class C { ${arg} }`, () => {
       t.doesNotThrow(() => {
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3742,7 +3742,7 @@ export function parseSuperExpression(
     case Token.LeftParen: {
       // The super property has to be within a class constructor
       if ((context & Context.SuperCall) < 1) report(parser, Errors.SuperNoConstructor);
-      if (context & Context.InClass) report(parser, Errors.UnexpectedPrivateField);
+      if (context & Context.InClass) report(parser, Errors.InvalidSuperProperty);
       parser.assignable = AssignmentKind.CannotAssign;
       break;
     }
@@ -3751,7 +3751,7 @@ export function parseSuperExpression(
       // new super() is never allowed.
       // super() is only allowed in derived constructor
       if ((context & Context.SuperProperty) < 1) report(parser, Errors.InvalidSuperProperty);
-      if (context & Context.InClass) report(parser, Errors.UnexpectedPrivateField);
+      if (context & Context.InClass) report(parser, Errors.InvalidSuperProperty);
       parser.assignable = AssignmentKind.Assignable;
       break;
     }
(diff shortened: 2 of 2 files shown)


3cp:
I am really uncomfortable with the removal of the two lines `| Context.InClass`, especially the second one in ClassField branch.

There is probably some missing test case for the ClassField branch (with some kind of super call? to test the InClass logic in parseSuperExpression).

test/parser/next/private_methods.ts
```diff
@@ -192,7 +192,9 @@ describe('Next - Private methods', () => {
     'foo() { this.#m, (() => this)().#m }',
     'foo() { this.#m, (() => this)().#m }',
     'foo() { this.#m, (() => this)().#m }',
-    'method() { super.#x(); }'
+    'method() { super.#x(); }',
+    '#method() { super.x(); }',
+    '#method() { super.#x(); }'
```
aladdin-add, line 197: just to double confirm:
```js
#method(){ super(); }
```
should be invalid.

aladdin-add: I see. there is already a case to test it: https://github.com/meriyah/meriyah/blob/6de707a0efb3053767deaa36b1ed6979b0d3f873/test/parser/next/private_methods.ts#L13

aladdin-add (approved):

3cp:
Any idea why the ClassField sets InClass before? I removed it because I cannot think of a test case requiring that.

Issue #214, opened 2022-03-31
es2022 support


aladdin-add:
- [x] * Class Fields ([Private instance methods and accessors](https://github.com/tc39/proposal-private-methods), [Class Public Instance Fields & Private Instance Fields](https://github.com/tc39/proposal-class-fields), [Static class fields and private static methods](https://github.com/tc39/proposal-static-class-features))
- [x] * [RegExp Match Indices](https://github.com/tc39/proposal-regexp-match-indices)
- [x] * [Top-level await](https://github.com/tc39/proposal-top-level-await)
- [x] * [Ergonomic brand checks for Private Fields](https://github.com/tc39/proposal-private-fields-in-in)
- [x] * [Class Static Block](https://github.com/tc39/proposal-class-static-block)
- [x] * Arbitrary module namespace identifier names[https://github.com/tc39/ecma262/pull/2154]

aladdin-add:
TLA has been supported,
class fields is still in next => need be enabled by default.

julianjensen:
I see that top-level `await` has been checked above. Not sure if it's still in the works or not but, just in case, I thought I'd post this here to make sure that this is an upcoming feature and not a bug. Currently, `for await` when used as a top-level `await` is failing on version 4.2.1.

```
import { parse } from 'meriyah';

const parseExample = ( name, source ) => {
    try
    {
        parse( source, { module: true, next: true });
        console.log( `PASS: ${name}` );
    }
    catch
    {
        console.warn( `FAIL: ${name}` );
    }
};

parseExample( `async function with 'for await'`, `async function x() { for await ( const a of b ) console.log( 'Works!' ); }` );
parseExample( `top-level 'await'`, `await topLevelAwaitFunction( 'Works' );` );
parseExample( `top-level 'for await'`, `for await ( const a of b ) console.log( 'Also breaks' );` );
```
The above outputs:
```
PASS: async function with 'for await'
PASS: top-level 'await'
FAIL: top-level 'for await'
```

3cp:
test262 class-fields-private-in ([Ergonomic brand checks for Private Fields](https://github.com/tc39/proposal-private-fields-in-in)) is supported.

PR #217, opened 2022-05-01
feat(lexer): support new RegExp Indices flag "d"

3cp:
Related to #214

### DIFF
--- a/src/lexer/regexp.ts
+++ b/src/lexer/regexp.ts
@@ -59,13 +59,14 @@ export function scanRegularExpression(parser: ParserState, context: Context): To
   const bodyEnd = parser.index - 1;
 
   const enum RegexFlags {
-    Empty = 0b000000,
-    IgnoreCase = 0b000001,
-    Global = 0b000010,
-    Multiline = 0b000100,
-    Unicode = 0b010000,
-    Sticky = 0b001000,
-    DotAll = 0b100000
+    Empty = 0b0000000,
+    IgnoreCase = 0b0000001,
+    Global = 0b0000010,
+    Multiline = 0b0000100,
+    Unicode = 0b0010000,
+    Sticky = 0b0001000,
+    DotAll = 0b0100000,
+    Indices = 0b1000000
   }
 
   let mask = RegexFlags.Empty;
@@ -91,7 +92,7 @@ export function scanRegularExpression(parser: ParserState, context: Context): To
         break;
 
       case Chars.LowerU:
-        if (mask & RegexFlags.Unicode) report(parser, Errors.DuplicateRegExpFlag, 'g');
+        if (mask & RegexFlags.Unicode) report(parser, Errors.DuplicateRegExpFlag, 'u');
         mask |= RegexFlags.Unicode;
         break;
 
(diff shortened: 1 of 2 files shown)


3cp:
Not behind `next` since it's a stage 4 feature.

test/lexer/regexp.ts
```diff
@@ -171,21 +173,19 @@ describe('Lexer - Regular expressions', () => {
       t.throws(() => scanSingleToken(state, context, 0));
     });
   }
-  fail('fails on /i/ii', '/ ', Context.AllowRegExp);
```
3cp, line 174: There are lots of wrong message in this sections. And there are many duplicated tests on `/i/yy` and `/i/ss`.

aladdin-add (approved):

src/lexer/regexp.ts
```diff
@@ -136,6 +142,8 @@ export function scanRegularExpression(parser: ParserState, context: Context): To
  */
 function validate(parser: ParserState, pattern: string, flags: string): RegExp | null | Token {
   try {
+    // Temporarily allows older version of Nodejs (or browser) to accept the new Indices flag "d".
+    flags = flags.replace('d', '');
```
aladdin-add, line 146: will it return a wrong value, even if the `/d` is supported?

3cp: Pls elaborate.

aladdin-add: e.g. `value: /./d` =>  `/./` (as the `'/d'` has been replaced)?

aladdin-add: ok. seems it won't happen (as stated in the valid test). :)

aladdin-add (approved):

3cp:
I overlooked the validation result is used.
```js
parser.tokenValue = validate(parser, pattern, flags);
```
So the tokenValue is wrong for regex with 'd' flag. Not sure what's the best solution. It might need a try catch to fall back to removing "d", so that at least for Nodejs v18, the tokenValue is correct.

aladdin-add:
a straightforward way is to do a check in the entry, something like:

```js
var isDsupported = true;
try {new Regexp('.', 'd');}catch(e){isDsupported = false;}
```

we may also want to check others like `/s`, `/u`, `/y`...

btw, I don't think it's a good resolution to just replace them to '', as they may be misused as normal values. e.g. `/./d` => `/./`. what do you think just return an exceptional value, e.g. `null`?

fisker:
Is this stored as RegExpLiteral value? It should be null per ESTree spec https://github.com/estree/estree/blob/master/es5.md#regexpliteral

3cp:
```
> r.body[0].expression
{ type: 'Literal', value: /a/g, regex: { pattern: 'a', flags: 'g' } }
```

The value is for parent interface just above your link.

```
interface Literal <: Expression {
    type: "Literal";
    value: string | boolean | null | number | RegExp;
}
```

3cp:
Oh, sorry. I got it now.

3cp:
What about
```js
try {
  return new RegExp(pattern, flags);
} catch (e) {
  try {
    // Fallback to only validate patterns on systems 
    // where some flags are not supported.
    new RegExp(pattern);
    return null;
  } catch (e) {
    report(parser, Errors.UnterminatedRegExp);
  }
}
```

aladdin-add:
it will allow unknown/dulp flags, e.g. `/foo/a`, `/foo/gg`?

3cp:
Validation happens after the duplication check.

PR #218, opened 2022-05-05
fix: use null as regex value in environment missing flag "d" support

3cp:
Continue of #217

### DIFF
--- a/src/lexer/regexp.ts
+++ b/src/lexer/regexp.ts
@@ -142,10 +142,15 @@ export function scanRegularExpression(parser: ParserState, context: Context): To
  */
 function validate(parser: ParserState, pattern: string, flags: string): RegExp | null | Token {
   try {
-    // Temporarily allows older version of Nodejs (or browser) to accept the new Indices flag "d".
-    flags = flags.replace('d', '');
     return new RegExp(pattern, flags);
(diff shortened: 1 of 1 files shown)


src/lexer/regexp.ts
```diff
@@ -142,10 +142,15 @@ export function scanRegularExpression(parser: ParserState, context: Context): To
...
 function validate(parser: ParserState, pattern: string, flags: string): RegExp | null | Token {
   try {
-    // Temporarily allows older version of Nodejs (or browser) to accept the new Indices flag "d".
-    flags = flags.replace('d', '');
     return new RegExp(pattern, flags);
   } catch (e) {
-    report(parser, Errors.UnterminatedRegExp);
+    try {
+      // Some JavaScript engine has not supported flag "d".
+      new RegExp(pattern, flags.replace('d', ''));
```
3cp, line 149: Could not just fallback to `new RegExp(pattern)`. Some flag affects the correctness of the pattern like `/\1/u` is incorrect, but `/\1/` is correct.

So we only anticipate missing "d" support here.

3cp:
Unknown flag and duplicated flags were checked before running validate().

aladdin-add (approved):
LGTM

PR #219, opened 2022-05-07
feat(parser): Add support for class static initialization block

### LINKED ISSUE #194: implement static blocks (stage 3)
Meriyah supports stage 3 proposals so need to implement static block

prantlf:
Attempts to fix #194 and one point from #214.

    extend interface ClassBody {
      body: [ MethodDefinition | PropertyDefinition | StaticBlock ];
    }

    interface StaticBlock <: BlockStatement {
      type: "StaticBlock";
    }

Proposal: https://github.com/tc39/proposal-class-static-block
MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Class_static_initialization_blocks
V8 examples: https://v8.dev/features/class-static-initializer-blocks
ESTree: https://github.com/estree/estree/blob/master/es2022.md#staticblock

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1771,6 +1771,41 @@ export function parseCatchBlock(
 }
 
 /**
+ * Parses class static initialization block
+ *
+ * @see [Link](https://github.com/tc39/proposal-class-static-block)
+ *
+ * @param parser Parser object
+ * @param context Context masks
+ * @param scope Scope instance
+ * @param start Start pos of node
+ * @param line
+ * @param column
+ */
+export function parseStaticBlock(
+  parser: ParserState,
+  context: Context,
+  scope: ScopeState | undefined,
+  start: number,
+  line: number,
+  column: number
+): ESTree.StaticBlock {
+  // StaticBlock ::
+  //   '{' StatementList '}'
+
+  if (scope) scope = addChildScope(scope, ScopeKind.Block);
+
+  const ctorContext = Context.InClass | Context.SuperCall;
+  context = ((context | ctorContext) ^ ctorContext) | Context.SuperProperty;
+  const { body } = parseBlock(parser, context, scope, {}, start, line, column);
+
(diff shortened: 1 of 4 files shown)


src/parser.ts
```diff
@@ -8302,6 +8337,8 @@ function parseClassElementList(
     key = parsePrivateIdentifier(parser, context | Context.InClass, tokenPos, linePos, colPos);
   } else if (context & Context.OptionsNext && (parser.token & Token.IsClassField) === Token.IsClassField) {
     kind |= PropertyKind.ClassField;
+  } else if (context & Context.OptionsNext && isStatic && token === Token.LeftBrace) {
```
3cp, line 8340: We don't need to check OptionsNext since this is a stage 4 feature.

prantlf: Oh yes, you're right. The time goes by so fast :-)

3cp (approved):
LGTM

3cp:
@aladdin-add could you have a look too?

test/parser/statements/static-block.ts
```diff
@@ -0,0 +1,143 @@
+import { Context } from '../../../src/common';
+import { pass, fail } from '../../test-utils';
+
+describe('Next - Class static initialization block', () => {
```
aladdin-add, line 4: good to have a few more tests. afaict, the test262 has many failing cases. https://github.com/tc39/test262/commit/afe217b318df5f197e64b30a8b5a4b391c777359

we can try add the test262 at some point, but surely it should be another PR, and won't block merging this one.

aladdin-add (approved):
LGTM, thanks!

3cp:
There are works to be done before we can ship it. I tried first few test262, it didn't work as expected.

3cp:
I don't have time to improve this for now. Just released v4.3.0 to include this feature.

Issue #224, opened 2022-08-19
object property encoding error


jdalton:
```js
/* Generated from: /test/language/expressions/assignment/member-expr-ident-name-break-escaped.js */
memberExprIdentNameBreakEscaped: {
  var obj = {};
  obj.bre\u0061k = 42;
  assert.sameValue(obj['break'], 42, 'property exists');
}
```
incorrectly throws:
Dot property must be an identifier

jdalton:
Thank you @3cp!

3cp:
Released in v4.3.1.

PR #225, opened 2022-09-10
Fix escaped async keyword and escaped dot property

### LINKED ISSUE #221: identifier encoding causes parse error
This is an entry from a test262 test:

```js
headLhsAsyncEscaped: {
  let async;
  for (\u0061sync of [7]);
  assert.sameValue(async, 7);
}
```
Meriyah incorrectly throws:
Unexpected token: 'escaped keyword'

3cp:
fix: async is not reserved word, and some fix on reserved words
fix: dot property should allow escaped keyword in strict mode
closes #221, #224

### DIFF
--- a/src/lexer/identifier.ts
+++ b/src/lexer/identifier.ts
@@ -85,13 +85,20 @@ export function scanIdentifierSlowCase(
     }
 
     if (context & Context.Strict) {
-      return token === Token.StaticKeyword
-        ? Token.EscapedFutureReserved
-        : (token & Token.FutureReserved) === Token.FutureReserved
-        ? Token.EscapedFutureReserved
-        : (token & Token.Reserved) === Token.Reserved
-        ? Token.EscapedReserved
-        : Token.AnyIdentifier;
+      if (token === Token.StaticKeyword) {
+        return Token.EscapedFutureReserved;
+      }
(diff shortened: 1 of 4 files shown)


3cp:
I am not very confident about my understanding of the spec and meriyah code.
This fix is only as good as our unit tests coverage :-)

3cp:
@aladdin-add do you have privilege to turn off (or delete) meriyah project on circleci? We don't use circelci anymore, CI runs on GitHub Actions now.

aladdin-add (approved):
LGTM~

aladdin-add:
closing and reopening to retrigger the ci.

fisker (approved):
PR looks good.

I'm not familiar with the spec, I'm not sure when it's a keyword. The JS parsing is strange..

I told people this `for (let of of let);` as joke.


PS: 你们两个中秋就不休息休息？

3cp:
JS introduced many new keywords but unfortunately has to be compatible with existing ancient legacy js code. It's probably the worst nightmare for any paser/tokenizer developer.

I would rather browser supports a clean `.js2` with no backward compatibility. The first two things I would like to have:
1. Get rid of undefined.
2. NaN equals NaN.

平时的时间都在赚钱，开源的事也只有在休息的时候有空做:-)

PR #227, opened 2022-10-01
fix(parser): ~ (0x7e) is a valid stop for Identifier or Keyword

### LINKED ISSUE #226: `for(x of~y);` is incorrectly rejected
Found this with [a fuzzer](https://github.com/shapesecurity/shift-fuzzer-js), so feel free to ignore; this didn't come up in real code.

Issue is as the title says: `for(x of~y);` is a legal program, though not one you'd normally encounter, and meriyah rejects it.

3cp:
closes #226


fisker:
Ah, so this character range is the solution to this https://github.com/vuejs/core/issues/6499, didn't know that. 👍

fisker (approved):

PR #232, opened 2022-12-18
fix(parser): "static" can be used as an identifier in ClassElement

### LINKED ISSUE #231: Class member named `static` doesn't parse correctly
```console
> require('meriyah').parse('class A {static;}', {next: true}).body[0].body.body[0]
{
  type: 'PropertyDefinition',
  key: null,
  value: null,
  static: true,
  computed: false,
  decorators: []
}
```

The key should be an `Identifier`.

3cp:
closes #231

Not 100% sure about the implementation.

The spec didn't tell me exact steps. https://262.ecma-international.org/13.0/#sec-static-semantics-isstatic

I tried to use an allowed-list (~~whitelist~~), but there are few conditions like `static "str-id" = 1;` `static *generator()`, I am not confident to cover them all.

So I extended the current blocked-list (~~blacklist~~) for the logic.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -8281,7 +8281,12 @@ function parseClassElementList(
 
     switch (token) {
       case Token.StaticKeyword:
-        if (!isStatic && parser.token !== Token.LeftParen) {
+        if (
+          !isStatic &&
+          parser.token !== Token.LeftParen &&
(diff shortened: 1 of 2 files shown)


fisker (approved):
I'm not familiar with parsers, but  LGTM. Thanks!

PR #238, opened 2023-03-13
Fix issue with import located in a sequence

ypapouski:
The PR fixes an error with the following construct:
```js
function() {
   import('some.js').then(e => e || {}), (this.some = !0);
}
```


3cp:
Thank you!

3cp (approved):

PR #242, opened 2023-05-09
Ensure that types are exported

diogoazevedos:
Fix for the following problem:

```
Could not find a declaration file for module 'meriyah'. '/path/to/node_modules/meriyah/dist/meriyah.esm.mjs' implicitly has an 'any' type.
  There are types at '/path/to/node_modules/meriyah/dist/src/meriyah.d.ts', but this result could not be resolved when respecting package.json "exports". The 'meriyah' library may need to update its package.json or typings.

import { parseScript } from 'meriyah';
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
   "unpkg": "dist/meriyah.umd.min.js",
   "exports": {
     "import": "./dist/meriyah.esm.mjs",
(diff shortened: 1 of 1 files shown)


3cp (approved):
Thx!

3cp:
Please try v4.3.6.

PR #246, opened 2023-05-10
Fix parsing of nested class with constructor

DimaIT:
Fixes the issue described below by remembering the HasConstructor flag and restoring it after parsing the class body.

---

Meriyah fails to parse the following code:

```
new class {
  start() {
    new class {
      constructor() {}
    }
  }

  constructor() {}          
}
```

Seems that the position of the outer constructor is what causes the problem cause when the outer constructor is above the start method Meriyah can parse it.

This snippet succeeds

```
new class {
  constructor() {}          

  start() {
    new class {
      constructor() {}
    }
  }
}
```

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -8204,6 +8204,8 @@ export function parseClassBody(
 
   consume(parser, context | Context.AllowRegExp, Token.LeftBrace);
   context = (context | Context.DisallowIn) ^ Context.DisallowIn;
+
+  let hasConstr = parser.flags & Flags.HasConstructor;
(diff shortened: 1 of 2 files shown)


3cp (approved):
Thx!

3cp:
It lets me to think: to deal with similar nested situations, a stack of flags/context might be better.

3cp:
There are some minor issues in unit tests to be fixed, missing `"decorators": []` in various places.

DimaIT:
> There are some minor issues in unit tests to be fixed, missing "decorators": [] in various places.

Fixed and added a couple more tests.

3cp:
Thx!
v4.3.7 is released.

PR #247, opened 2023-10-31
Fix assignment as a value for property definition

DimaIT:
Double assignment is not parsed correctly in class' property definition.

`class A { b = c = d; }` is parsed like `class A { b = c; = d; }` which is not a valid AST.

This PR fixes it along with other possible assignment operators.


3cp (approved):
Thank you!

3cp:
v4.3.8 is released.

PR #249, opened 2023-11-23
fix for regexp test after do-while with semicolon

lauraexplora:
Fix for bug found when attempting to parse a regexp test immediately after a do-while loop terminated by a semicolon.


3cp:
Thank you!

3cp:
I didn't realize this one merges to outdated dev branch. We have not used that branch for long time. Can you raise another one for master branch? Thanks!

PR #252, opened 2024-03-04
feat(parser): add onInsertedSemicolon option

dmichon-msft:
Adds support for the `onInsertedSemicolon` callback option, matching the base behavior of the `acorn` option of the same name. This option is used by `webpack` to identify when replacing the text of a node with a parenthesized expression would run afoul of a difference in runtime behavior if automatic semicolon insertion occurred.

Motivating example:
```js
console.log('hi!') // ASI occurs here
foo() // ASI occurs here
console.log(foo.name)
```

If webpack replaces `foo` with the imported expression `(0,some_module.foo)`, the resulting code has the runtime behavior of

```js
console.log('hi!')(0,some_module.foo)() // tried to call the result of `console.log`, TypeError
console.log((0,some_module.foo).name) // Correct code
```

On the other hand, if webpack tries to always insert a semicolon, we instead have

```js
console.log('hi') // ASI not needed
;(0,some_module.foo)() // Correct code
console.log(;(0,some_module.foo).name) // Syntax error
```

Ultimately this is an issue because webpack doesn't perform all transforms on the AST layer and run the code through a printer for performance reasons.

The lack of this option is the one difference in behavior between `acorn` and `meriyah` that would potentially impact using `meriyah` in `webpack` for a 20-30% AST parse performance boost.

Adds a basic unit test for the feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,9 @@ This is the available options:
   // Allows comment extraction. Accepts either a function or array
   onComment: []
 
+  // Allows detection of automatic semicolon insertion. Accepts a callback function that will be passed the charater offset where the semicolon was inserted
+  onInsertedSemicolon: (pos) => {}
+
   // Allows token extraction. Accepts either a function or array
   onToken: []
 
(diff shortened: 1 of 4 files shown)


3cp:
The change looks trivial! Glad it's easy to meet your requirement.

3cp (approved):

3cp:
v4.4.0 is released.

PR #254, opened 2024-03-25
Fix usage of `async` and `import` in comma expression

DimaIT:
Fixes usage of `async` and `import` in comma expressions, like:
```js
async = 1, b = 2;
b = 2, async = 1;
import.meta, 1;
```

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1119,7 +1119,9 @@ export function parseAsyncArrowOrAsyncFunctionDeclaration(
    * https://github.com/estree/estree/blob/master/es5.md#sequenceexpression
    *
    */
-  if (parser.token === Token.Comma) expr = parseSequenceExpression(parser, context, 0, start, line, column, expr);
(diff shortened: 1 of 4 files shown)


src/parser.ts
```diff
@@ -3574,6 +3581,8 @@ export function parseAsyncExpression(
     if (inNew) report(parser, Errors.InvalidAsyncArrow);
     return parseArrowFromIdentifier(parser, context, parser.tokenValue, expr, inNew, canAssign, 0, start, line, column);
   }
+
+  parser.assignable = AssignmentKind.Assignable;
```
3cp, line 3585: Given your change on line 1114, should this line be added before line 2694?

DimaIT: It seems not needed here, added a couple more test-cases to cover it

3cp (approved):

3cp:
v4.4.1 is released.

PR #256, opened 2024-05-22
Allow `super` to be used in arrow functions and in static object initialisation

ypapouski:
The PR fixes issues described in this issue: https://github.com/meriyah/meriyah/issues/257

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3931,11 +3931,7 @@ export function parseMemberOrUpdateExpression(
       case Token.Period: {
         nextToken(parser, (context | Context.AllowEscapedKeyword | Context.InGlobal) ^ Context.InGlobal);
 
-        if (
-          context & Context.InClass &&
-          (parser.token & Token.PrivateField) === Token.PrivateField &&
-          parser.tokenValue === 'super'
-        ) {
+        if (context & Context.InClass && parser.token === Token.PrivateField && parser.tokenValue === 'super') {
           report(parser, Errors.InvalidSuperProperty);
         }
 
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3803,7 +3803,9 @@ export function parseSuperExpression(
     case Token.LeftParen: {
       // The super property has to be within a class constructor
       if ((context & Context.SuperCall) === 0) report(parser, Errors.SuperNoConstructor);
-      if (context & Context.InClass) report(parser, Errors.InvalidSuperProperty);
+      if (context & Context.InClass && !(context & Context.InMethod)) {
+        report(parser, Errors.InvalidSuperProperty);
+      }
       parser.assignable = AssignmentKind.CannotAssign;
       break;
     }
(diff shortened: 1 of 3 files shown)


src/parser.ts
```diff
@@ -3927,6 +3931,14 @@ export function parseMemberOrUpdateExpression(
       case Token.Period: {
         nextToken(parser, (context | Context.AllowEscapedKeyword | Context.InGlobal) ^ Context.InGlobal);
 
+        if (
+          context & Context.InClass &&
+          (parser.token & Token.PrivateField) === Token.PrivateField &&
```
3cp, line 3936: I am confused, why not do `parser.token === Token.PrivateField`?

src/parser.ts
```diff
@@ -8531,6 +8552,16 @@ export function parsePropertyDefinition(
 
     if (parser.token === Token.Arguments) report(parser, Errors.StrictEvalArguments);
 
+    const modifierFlags =
```
3cp, line 8555: I believe you copied some code from parseMethodDefinition.
There are few places that meriyah uses this hard to cipher binary, Kenny was very focused on performance, but he is retired now. I could not remember what it does (if I ever understood it before). Could you help to add few comments for what is does?
Thanks very much!

src/parser.ts
ypapouski: Well, based on how `parser.token` is used in other places I made an assumption that `parser.token` might contain multiple values combined together and therefore bitwise AND is required, for example the same method does similar comparison located at the very beginning of it:
```js
if ((parser.token & Token.IsUpdateOp) === Token.IsUpdateOp && (parser.flags & Flags.NewLine) === 0) {
    expr = parseUpdateExpression(parser, context, expr, start, line, column);
  } else if ((parser.token & Token.IsMemberOrCallExpression) === Token.IsMemberOrCallExpression) {
...
```

src/parser.ts
ypapouski: Yes, you are right, I copied it from another place, namely from the `parseMethodDefinition`. Why I did it. It's because I noticed that Meriyah didn't throw the exception for the following js `class a { a() { super.a }}` and apparently it's because `parseMethodDefinition` changes the context. So, I did the same for `parsePropertyDefinition` and it started working for `class a { a = () => { super.a }}`.

src/parser.ts
3cp: IsUpdateOp is different, it's not a token, but a trait shared by few real token types. Those traits are using higher bits (bit 13 to 31).

```js
  Assign                  = 29 | IsAssignOp | IsClassField, // =
  ShiftLeftAssign         = 30 | IsAssignOp, // <<=
  ShiftRightAssign        = 31 | IsAssignOp, // >>=
  LogicalShiftRightAssign = 32 | IsAssignOp, // >>>=
```

PrivateField doesn't turn on any of those traits, 131 is well below bit 13.
```js
PrivateField      = 131,
```

src/parser.ts
3cp: OK :-) Leave it then.

src/parser.ts
ypapouski: Ah, I see your point! 👍 Thanks, I will correct then!

ypapouski: Changed in a056767af5ad0ec8124ca0bc988dc4fd32012fe3

3cp (approved):

3cp:
v4.4.3 is released.

PR #261, opened 2024-06-05
chore: remove option identifierPattern

3cp:
This option added a "pattern" flag to Identifer node, but this feature was rejected by ESTree, https://github.com/estree/estree/issues/196 KFlash prematurelly added this when reading it.
Since it's not in ESTree spec, nobody would never relied on it. In this sense, not a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,9 +100,6 @@ This is the available options:
   // Adds a source attribute in every node’s loc object when the locations option is `true`
   source: false;
 
-  // Distinguish Identifier from IdentifierPattern
-  identifierPattern: false;
-
    // Enable React JSX parsing
   jsx: false
 }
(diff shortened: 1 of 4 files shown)


fisker (approved):
Make sense!

fisker:
> At least 1 approving review is required by reviewers with write access.

I don't have write access, so my approval won't unblock.

3cp:
> > At least 1 approving review is required by reviewers with write access.
> 
> I don't have write access, so my approval won't unblock.

Thx! That confused hell out of me. Write permission assigned.

fisker:
> Write permission assigned.

I don't know that I'm a member here..

> That confused hell out of me. 

The check mark was gray (not green) before

PR #265, opened 2024-06-06
Remove unused `isArrow` from `classifyIdentifier`

fisker:
I don't really understand what is that for, but found it's always `1`, and rollup is able to remove it during #264.


3cp (approved):

3cp:
Probably some left over code.

fisker:
Should we consider use squash merging to keep the history clean? We can uncheck this

image

3cp:
Sure.

PR #266, opened 2024-06-06
chore: show proper error "Expected 'from'"

3cp:
There is an existing method does report better error.

```js
export function consume(parser: ParserState, context: Context, t: Token): void {
  if (parser.token !== t) report(parser, Errors.ExpectedToken, KeywordDescTable[t & Token.Type]);
  nextToken(parser, context);
}
```


fisker:
No tests?

3cp:
Oh yes. I will add some.

3cp:
Just realised we have never tested error message before in the whole unit tests:(
All previous tests just tested there is error thrown.

fisker (approved):

PR #270, opened 2024-06-07
chore: update TypeScript to v5

fisker:


src/common.ts
```diff
@@ -831,3 +831,9 @@ export function classifyIdentifier(parser: ParserState, context: Context, t: Tok
 
   if (!isValidIdentifier(context, t)) report(parser, Errors.Unexpected);
 }
+
+/**
+ * A helper function to check parser state, workaround TS2678 error
+ * @see https://github.com/microsoft/TypeScript/issues/9998
+ */
+export const is = (value: number | boolean) => value;
```
fisker, line 839: It's annoying that TS narrowing down the type of `parser.token` incorrectly, [issue](https://github.com/microsoft/TypeScript/issues/9998), this function prevents that.

Do you have a better idea?

src/parser.ts
```diff
@@ -5528,7 +5533,7 @@ function parseSpreadOrRestElement(
     );
 
     const { token, tokenPos, linePos, colPos } = parser;
-
+    // @ts-expect-error BUG?
     if (token === Token.Assign && token !== closingToken && token !== Token.Comma) {
```
fisker, line 5537: Is this a mistake?

src/common.ts
3cp: I will have a play.

src/parser.ts
3cp: line 5485 `if (token !== Token.Assign && token !== closingToken && token !== Token.Comma) {`
That's more like it.

I tested this line with `if (token === Token.Assign) {`, all tests still pass. Please fix the code. Thx!

src/common.ts
3cp: Is this a tsc v5 bug? tsc v4 has no such issue, but tsc v4 fails all the places.

3cp: I would suggest to hold off tsc v5 before we have a proper solution.

src/parser.ts
fisker: But I use `if (token === Token.Assign)`, test also pass.

3cp: #271

3cp:
I have seen weird ts v5.5 typing issue in other projects.
ts v5 seems trying to be too smart.

fisker:
Maybe this solution is acceptable? https://github.com/microsoft/TypeScript/issues/9998#issuecomment-235963457

3cp:
I think it's a ts bug, it assumes our `nextToken(parser, ...)` didn't mutate `parser.token`.
But `parser.token =` is right there in the implementation of `nextToken`.

3cp:
I can reproduce the issue https://www.typescriptlang.org/play/?ts=5.5.3#code/JYOwLgpgTgZghgYwgAgCoQM5mQbwFDKHIwA2cA5gFzIAMyAPsgIwDceAvnnjAK4gJhgAexDIAtjzBxIACkhZq6LAEpcBIvLAA6UhWQBeZm054EIrMjiLM2QzmJkqtZOzZ5gMZDLg7HB-YY0qvhE4pLSEN7KbKEeXj665P6GTMHqoYRmIBhCJBBaJELkMgDkiSUANJa+FNHpLhx4QA

But the same typing error also pops up with ts v4.

3cp:
Had a read of the ts github issue. It sounds like ts would not address this issue anytime soon.

If you have time to apply the `parser.currentToken() or getToken()` solution, please go ahead :-)

3cp:
I am also very confused why we didn't have this issue with ts v4 and v3 before.

fisker:
Done.

3cp (approved):

Issue #274, opened 2024-07-17
Weird context mutation


3cp:
https://github.com/meriyah/meriyah/blob/434279f78e90a6a8d4e5d891fc3b2b8414030c36/src/parser.ts#L4836

The two binary bits used there are Context.Strict and Context.Module, but the `<< 11` doesn't make sense here.

I tried to remove the whole thing, change that line to `validateFunctionName(parser, context, parser.token);`. All unit tests still pass.

@fisker could you help to have a look? I don't know if missed anything here.

3cp:
```

    context =
      ((context | modifierFlags) ^ modifierFlags) |
      ((state & 0b0000000000000000000_0000_01011000) << 18) |
      0b0000110000001000000_0000_00000000;
```

Found more, the `<< 18` doesn't make any sense either. the 01011000 is OptionsDirectives, OptionsJSX, OptionsLexical.
Left shifting 18 of them will become InYieldContext, InAwaitContext, InConstructor.

3cp:
> https://github.com/meriyah/meriyah/blob/434279f78e90a6a8d4e5d891fc3b2b8414030c36/src/parser.ts#L4836
> 
> The two binary bits used there are Context.Strict and Context.Module, but the `<< 11` doesn't make sense here.
> 
> I tried to remove the whole thing, change that line to `validateFunctionName(parser, context, parser.token);`. All unit tests still pass.
> 
> @fisker could you help to have a look? I don't know if missed anything here.

Got it, the code tried to turn on Context.InYieldContext when Context.Strict, and Context.InAwaitContext when Context.Module. But the logic is not needed now, because validateFunctionName checks `(Context.InAwaitContext | Context.Module)` and ` (Context.InYieldContext | Context.Strict)`.

fisker:
小黄鸭调试法。

PR #275, opened 2024-07-17
chore: remove explicit binary mask

3cp:
This removes the coupling with the Context and PropertyKind enum definition.
The old code means we cannot modify Context and PropertyKind definition freely.

Related to #274


fisker (approved):

3cp:
One more change.

fisker (approved):

PR #278, opened 2024-07-18
feat(parser): class public/private fields are in ecma spec now

3cp:
BREAKING CHANGE: move class public/private fields out of options.next

Also fixed a bug that prevented rejecting "class A { get x }".


test/test262-parser-tests/parser-tests.ts
```diff
@@ -19,6 +19,10 @@ const skipped = {
     '79f882da06f88c9f.js',
     '92b6af54adef3624.js',
 
+    // class fields
+    '98204d734f8c72b3.js',
+    'ef81b93cf9bdb4ec.js',
```
3cp, line 24: test262-parser-tests seems quite out of date. The two files contain perfect valid JS code with class property definition.

```js
// 98204d734f8c72b3.js
(class {a})
// ef81b93cf9bdb4ec.js
(class {a=0})
```

3cp:
@fisker do you know the relation between [test262-parser-tests](https://github.com/tc39/test262-parser-tests) and [test262](https://github.com/tc39/test262)?

The test262-parser-tests we use seems quite out of date.

3cp:
Our `options.next` is seriously behind current ecmascript spec. This is a beginning PR that I want to move some features out of `next` flag.

src/parser.ts
```diff
@@ -8620,14 +8620,29 @@ export function parsePropertyDefinition(
...
+          decorators
+        }
+      : {
+          type: 'PropertyDefinition',
+          key,
+          value,
+          static: (state & PropertyKind.Static) > 0,
+          computed: (state & PropertyKind.Computed) > 0
+        }) as any
+  );
```
fisker, line 8645: These two branches are quite big, can we use 

```js
decorator: (context & Context.OptionsNext ? decorator : undefined)
```

?

Or

```js
...(context & Context.OptionsNext ? {decorator} : undefined)
``

if tge key existence matters.

fisker:
> @fisker do you know the relation between [test262-parser-tests](https://github.com/tc39/test262-parser-tests) and [test262](https://github.com/tc39/test262)?
> 
> The test262-parser-tests we use seems quite out of date.

Didn't involved in parser much, I'm not familiar with them.

src/parser.ts
3cp: There are few cases like this. I will clean them up.

fisker (approved):

PR #279, opened 2024-07-19
test: update test262 parser test

fisker:
Just copied from [acorn](https://github.com/acornjs/acorn/blob/12ad1164a70dbe804f1ece822f34d100545f1afa/bin/run_test262.js)


fisker:
It takes me a while, but we do have lots of tests can't pass in class fields/methods.

test/test262-parser-tests/unsupported-features.txt
```diff
@@ -0,0 +1,9 @@
+explicit-resource-management
+regexp-modifiers
+import-assertions
+import-attributes
+regexp-duplicate-named-groups
+regexp-v-flag
+arbitrary-module-namespace-names
+source-phase-imports
+source-phase-imports-module-source
```
fisker, line 9: @3cp Can you confirm this is the correct list of features that we don't support?

3cp: Regexp modifiers should work.
Meriyah doesn't parse regexp internal, the whole regex is one tokenValue, parsed by native RegExp provided by the runtime (which runs meriyah), `try { return new RegExp(...) } catch ...`.

In same sense, regexp-duplicate-named-groups properly also works.

3cp:
Don't know what's wrong with mocha in nodejs v22 on CI, I don't have the issue locally.

3cp:
The whitelist is ones we have trouble with?

test/test262-parser-tests/unsupported-features.txt
3cp: class-static-block is not supported.

3cp: class-static-methods-private is not supported.

fisker:
> Don't know what's wrong with mocha in nodejs v22 on CI, I don't have the issue locally.

https://github.com/nodejs/node/issues/53902



> The whitelist is ones we have trouble with?

Yes.

test/test262-parser-tests/unsupported-features.txt
fisker: > Meriyah doesn't parse regexp internal

I didn't know that at first, I think it's really bad to rely on runtime behavior, maybe we should do something in future.

3cp (approved):

Issue #253, opened 2024-03-19
Add `ImportAttribute` for import attributes (stage 3 Estree)


aralroca:
Now that runtimes like Bun supports macros and the proposal is in stage 3 it would be nice to support import attributes.

https://github.com/tc39/proposal-import-attributes

Example in TypeScript:
<img width="1724" alt="Screenshot 2024-03-19 at 22 14 21" src="https://github.com/meriyah/meriyah/assets/13313058/75c12ed7-8dd4-4ca0-be01-56fdd1c67ce8">

3cp:
## Note for implementation

https://github.com/estree/estree/blob/master/stage3/import-attributes.md

EStree which meriyah implements has different definition of this syntax.

3cp:
BTW I don't think the typescript screenshot is a correct reference :-)

The old (or current) typescript parser thinks that's the legacy with statement, not part of the import declaration. You can see they appear as two sibling nodes in the syntax tree.

aralroca:
It will be great to replace the TypeScript parser that I used in this plugin with Meriyah, which is much faster! 

https://github.com/aralroca/prerender-macro

3cp:
I cannot guarantee a time frame for this feature as I don't have enough bandwidth on open source projects right now. But I will put it on my TODO list :)

3cp:
Bun has a native parser. Maybe you can talk to them to expose some APIs. It's probably not in extree syntax, but nothing beats Bun's performance.

> the parser in Bun is a Zig port of esbuild's parser.

aralroca:
> Bun has a native parser. Maybe you can talk to them to expose some APIs. It's probably not in extree syntax, but nothing beats Bun's performance.
> 
> > the parser in Bun is a Zig port of esbuild's parser.

Yes, I have discussed it several times on discord but I really don't know if they have plans to do it in the short term. Maybe @Jarred-Sumner can confirm

aralroca:
@3cp Is there any news on this? 😊 if I can help in contributing I am open although I am not familiar with it.

3cp:
Sorry I have not put on any effort. You are more than welcome to try.

aralroca:
Maybe I'll take some time this coming week, any tips @3cp? I see that there is no CONTRIBUTING.md in the repo

3cp:
https://262.ecma-international.org/

Ecmascript spec defined all logic that a parser needs to follow. But it's a hard piece to digest.

[...]

PR #280, opened 2024-07-20
feat: support import attributes with `ImportAttribute`

aralroca:
Fixes https://github.com/meriyah/meriyah/issues/253

This PR implements support for Import Attributes, aligning with the [Import Attributes proposal](https://github.com/tc39/proposal-import-attributes) and [Estree import attributes](https://github.com/estree/estree/blob/master/stage3/import-attributes.md).


3cp:
Great work!

From your implementation I noticed following syntax will work:
```js
// trailing comma
import foo from "foo" with { type: "json", };
import("foo", {type: "module", });

// and some expression features like following.
import("foo", {[type = "a"]: "module"});
import("foo", {type: () => {}});
// some expression features can be rejected by playing Context and canAssign and inGroup.
```

I am not clear whether the proposal allows them, the proposal didn't have a detailed parsing algorithm like normal ecmascript spec provided. But we can probably leave with them for now. Suggest to add few more test lines to the pass/fail list.

Another thing is you need to remove `import-assertions` and `import-attributes` from `test/test262-parser-tests/unsupported-features.txt`. That will turn some more test262 test cases, sure they will give you some more work to do :-) For some you cannot figure out solution, the last thing we can do is to add some cases to `test/test262-parser-tests/whitelist.txt`. The whitelist are the cases meriyah has issue with, we do have a huge list of problems.

Happy to merge your work after passing (most :-)) test262 cases.

3cp:
Ignore the nodejs v22 failure. v22.5.0 has some issue right now.

3cp:
Many are in the whitelist right now, you need to remove them too.

3cp:
@fisker please help.
For this branch, I removed `import-assertions` and `import-attributes` from `unsupported-features.txt`, and various entries in `whitelist.txt`.

But the test262 summary didn't change anything before and after my edit?? Any other thing is controlling test262 cases?

```
huocp@Chunpengs-MacBook-Pro ~/p/meriyah (master)> bun run test
$ mocha --"test/**/*.ts"

  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬]Testing complete.
Summary:
 ✔ 81974 valid programs parsed without error
 ✔ 7479 invalid programs produced a parsing error
 ✔ 454 invalid programs did not produce a parsing error (and allowed by the whitelist file)
 ✔ 1843 valid programs produced a parsing error (and allowed by the whitelist file)
 ✔ 1674 programs were skipped
  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬]

  93947 passing (12s)

huocp@Chunpengs-MacBook-Pro ~/p/meriyah (master)> bun run test
$ mocha --"test/**/*.ts"

  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬]Testing complete.
Summary:
 ✔ 81974 valid programs parsed without error
 ✔ 7479 invalid programs produced a parsing error
 ✔ 454 invalid programs did not produce a parsing error (and allowed by the whitelist file)
 ✔ 1843 valid programs produced a parsing error (and allowed by the whitelist file)
 ✔ 1674 programs were skipped
  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬]

  93947 passing (12s)
```

3cp:
@fisker ignore my deleted comment. I was editing wrong directory.

3cp:
There are some valid reason we need some cases in whitelist.

Like `language/import/import-attributes/json-invalid.js`, which asks the parser to read and check the JSON file. Meriyah doesn't do that and would not do that.

```js
import value from './json-invalid_FIXTURE.json' with { type: 'json' };
```

3cp:
Correction: import-assertions is the outdated proposal that import-attributes replaced. Please keep import-assertions in unsupported list.

aralroca:
> There are some valid reasons we need some cases in whitelist.
> 
> Like `language/import/import-attributes/json-invalid.js`, which asks the parser to read and check the JSON file. Meriyah doesn't do that and would not do that.
> 
> ```js
> import value from './json-invalid_FIXTURE.json' with { type: 'json' };
> ```

This is the only case I added to the whitelist, the other ones are fixed! Feel free to re-review @3cp ! Thanks a lot!

src/parser.ts
```diff
@@ -4445,13 +4457,105 @@ export function parseImportExpression(
...
+    let options: ESTree.Expression | null = null;
+
+    if (parser.getToken() === Token.Comma) {
+      consume(parser, context, Token.Comma);
+
+      while (parser.getToken() !== Token.RightParen && parser.getToken() !== Token.Comma) {
+        if (parser.getToken() === Token.InKeyword) {
+          consume(parser, context, Token.InKeyword);
+        }
+        options = parseExpression(parser, context, 1, inGroup, parser.tokenPos, parser.linePos, parser.colPos);
```
aralroca, line 4475: I'm not sure about this, they pass the tests, but I was really just looking for “no crash”, but I think if I add tests of the expected AST to test262/test/language/expressions/dynamic-import/import-attributes/2nd-param-in.js it will crash. Tomorrow I will continue and add more tests from test262 with the AST after parsing.

src/parser.ts
```diff
@@ -4468,14 +4468,20 @@ export function parseImportExpression(
...
-      } else {
-        report(parser, Errors.InvalidImportExpression);
       }
     }
 
     node.options = options;
+
+    // Allow trailing comma after options
+    if (parser.getToken() === Token.Comma) {
+      consume(parser, context, Token.Comma);
```
3cp, line 4483: consumeOpt is for this kind of optional.

src/parser.ts
3cp: Yes, this is not right.
"in" is handled by parseBinaryExpression. The reason that parseExpression didn't cover it, is probably Context.DisallowIn is turned on. Probably need just `(context | Context.DisallowIn) ^ Context.DisallowIn` to turn it off.

aralroca: Okay perfect, it looks better now. I've added the failing test with the AST and corrected it with what you told me.

src/parser.ts
```diff
@@ -4445,13 +4457,99 @@ export function parseImportExpression(
...
+  while (parser.getToken() !== Token.RightBrace) {
+    const start = parser.tokenPos;
+    const line = parser.linePos;
+    const column = parser.colPos;
+
+    const key = parseIdentifierOrLiteral(parser, context);
+    consume(parser, context, Token.Colon);
+    const value = parseLiteral(parser, context);
+    const keyContent = key.type === 'Literal' ? key.value : key.name;
+    const isJSONImportAttribute = keyContent === 'type' && value.value === 'json';
```
fisker, line 4519: Why is this needed? The value of type shouldn't matters at all.

fisker: Ah, because we support `json-modules` feature? https://github.com/tc39/proposal-json-modules#why-dont-json-modules-support-named-exports

aralroca: @fisker this is to fix this test: `test262/test/language/import/import-attributes/json-named-bindings.js`. As it says:

```
In the early design of JSON modules, contributors considered allowing the
  properties of object values in JSON modules to be imported directly by name.
  This was ultimately rejected, so attempting to import in this way should
  produce a SyntaxError.
```

However, this is specific only to JSON. For other import attributes, such as Bun macros, it has to be done for named bindings: https://bun.sh/docs/bundler/macros

fisker: I don't think any parser rejects this now, maybe we can put `json-modules` into [unsupported features](https://github.com/meriyah/meriyah/blob/master/test/test262-parser-tests/unsupported-features.txt). But lets hear what @3cp thinks first.

aralroca: No problem, or I imagine it will be putting this `test262/test/language/import/import-attributes/json-named-bindings.js` to the whitelist. Well, I'll wait for @3cp's answer and see what he thinks. Thanks!

src/errors.ts
```diff
@@ -138,6 +138,8 @@ export const enum Errors {
   InvalidNestedStatement,
   UnknownLabel,
   InvalidImportTail,
+  InvalidImportExpression,
+  InvalidImportBinding,
```
3cp, line 142: Just noticed you forgot to add errorMessages for these two.

3cp: Which causes:
```
> parseModule('import { foo } from "./foo.json" with { type: "json" };', { next: true })
/Users/huocp/playground/meriyah/src/errors.ts:176
        const message = '[' + line + ':' + column + ']: ' + exports.errorMessages[type].replace(/%(\d+)/g, (_, i) => params[i]);
                                                                                        ^

Uncaught TypeError: Cannot read properties of undefined (reading 'replace')
```

src/parser.ts
```diff
@@ -4445,13 +4457,99 @@ export function parseImportExpression(
...
+  context: Context,
+  specifiers: ESTree.ImportDeclaration['specifiers'] = []
+): ESTree.ImportAttribute[] {
+  consume(parser, context, Token.WithKeyword);
+  consume(parser, context, Token.LeftBrace);
+
+  const attributes: ESTree.ImportAttribute[] = [];
+  const keysContent = new Set<ESTree.Literal['value'] | ESTree.Identifier['name']>();
+  const validJSONImportAttributeBindings =
+    specifiers.length >= 1 &&
```
3cp, line 4505: This enables
```js
import { default as a, foo } from "./foo.json" with { type: "json" };
// and
import a, { foo } from "./foo.json" with { type: "json" };
```

It probably should be `specifiers.length === 1`.

3cp: I think test262 missed some test cases for this.

src/parser.ts
3cp: I wish tc39 pushed this as runtime error, not syntax error :-)

src/parser.ts
```diff
@@ -4445,13 +4457,99 @@ export function parseImportExpression(
...
+    const column = parser.colPos;
+
+    const key = parseIdentifierOrLiteral(parser, context);
+    consume(parser, context, Token.Colon);
+    const value = parseLiteral(parser, context);
+    const keyContent = key.type === 'Literal' ? key.value : key.name;
+    const isJSONImportAttribute = keyContent === 'type' && value.value === 'json';
+
+    if (isJSONImportAttribute && !validJSONImportAttributeBindings) {
+      report(parser, Errors.InvalidImportBinding);
```
3cp, line 4522: The validJSONImportAttributeBindings logic can probably be delayed here. So we don't have the unnecessary cost for non json import.

src/parser.ts
```diff
@@ -4445,13 +4457,99 @@ export function parseImportExpression(
...
+    if (parser.getToken() !== Token.RightBrace) {
+      consume(parser, context, Token.Comma);
+    }
+  }
+
+  consume(parser, context, Token.RightBrace);
+  return attributes;
+}
+
+function parseIdentifierOrLiteral(parser: ParserState, context: Context): ESTree.Identifier | ESTree.Literal {
```
3cp, line 4547: Suggest to rename it to parseIdentifierOrStringLiteral.
Do we have any test cases to reject?
```
with { 1: "foo" }
with { type: 1 }
with { type: [1] }
with { type: null }
with { type: undefined }
with { type: 'json', foo: {} }
...
```

3cp (changes requested):

src/errors.ts
3cp: Should it be named InvalidJSONImportBinding?

src/parser.ts
3cp: Looks like your implementation didn't reject the above cases.

From the proposal: 
```
We could permit import attributes to have more complex values than simply strings, for example:

import value from "module" with { attr: { key1: "value1", key2: [1, 2, 3] } };
This would allow import attributes to scale to support a larger variety of metadata.

We propose to omit this generalization in the initial proposal, as a key/value list of strings already affords significant flexibility to start, but we're open to a follow-on proposal providing this kind of generalization.
```

3cp: test262 cases are way inadequate :-)

src/errors.ts
aralroca: I have made the changes that you have proposed, I have also deleted the InvalidImportExpression, since in the end it is not used after the changes to adapt to test252.

aralroca:
@3cp @fisker Thank you for all your suggestions for changes. Feel free to check the PR again, anything else feel free to say. Looking forward to have the import attributes supported 😊

3cp (approved):

3cp:
LGTM.
Thx!
Let @fisker have another look.

fisker (approved):
LGTM, thank you  @aralroca

src/parser.ts
```diff
@@ -4445,15 +4451,79 @@ export function parseImportExpression(
...
   const source = parseExpression(parser, context, 1, inGroup, parser.tokenPos, parser.linePos, parser.colPos);
+  let options: ESTree.Expression | null = null;
+
+  if (parser.getToken() === Token.Comma) {
+    consume(parser, context, Token.Comma);
+
+    if (parser.getToken() === Token.LeftBrace) {
+      options = parseExpression(parser, context, 1, inGroup, parser.tokenPos, parser.linePos, parser.colPos);
+
+      if (options && options.type !== 'ObjectExpression') {
```
Tcjmd: In this case can it become a binding? Anyway the expr parser should detect and throw if missing RightBrace. Meaning this check is both expensive due to perf and not needed

src/parser.ts
```diff
@@ -4445,15 +4451,79 @@ export function parseImportExpression(
...
+    const line = parser.linePos;
+    const column = parser.colPos;
+
+    const key = parseIdentifierOrLiteral(parser, context);
+    consume(parser, context, Token.Colon);
+    const value = parseLiteral(parser, context);
+
+    attributes.push(
+      finishNode(parser, context, start, line, column, {
+        type: 'ImportAttribute',
```
Tcjmd, line 4533: This is handcoded and no AST node. No ESTree node for this?

3cp: Most finishNode usage in meriyah looks like this, with plain string type.

src/parser.ts
3cp: That's not latest code.

Tcjmd: My mistake. Can you link to latest. On cp

3cp: Just visit the changed code in this PR, you will see latest code.

PR #283, opened 2024-07-25
feat(parser): move hashbang out of "next" option

3cp:
Hashbang is a finished proposal. https://github.com/tc39/proposals/blob/main/finished-proposals.md

BREAKING CHANGE: hashbang now works without "next" option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,10 +29,8 @@
 ## ESNext features
 
 - [Decorators](https://github.com/tc39/proposal-decorators)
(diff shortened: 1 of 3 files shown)


3cp:
I think the "next" option is in now sync with spec. We only have import-attributes and decorators behind "next" right now. Plus json-modules is also supported by import-attributes.

All other stage-3 proposals are not implemented yet, or not for a parser.

fisker (approved):

Issue #236, opened 2023-02-24
Reduce bundle size by using `meriyah.esm.mjs` in favor of `meriyah.esm.js`


benmccann:
Right now there is both a `meriyah.esm.js` and `meriyah.esm.mjs` that are distributed. These appear to be the same file:

https://github.com/meriyah/meriyah/blob/15f426f9871e923ec7396b5e35a33b976ed6c5c3/scripts/bundle.js#L18

Distributing the same file with two different names makes the npm bundle larger whereas if `meriyah.esm.mjs` were used exclusively the bundle size would be smaller

3cp:
When we enabled cjs and mjs files, the js files were kept for backwards compatibility in case users have explicitly used them.

I agree we need to clean the dist files up. But we prefer doing it together with a major version release, that would also allow us to remove more outdated files in current JS lands. For example the systemjs format files.

fisker:
Fixed by #284

PR #284, opened 2024-07-25
chore: reduce dist files to esm, cjs, and umd

3cp:
BREAKING CHANGE: only distribute following files:
dist/meriyah.cjs
dist/meriyah.mjs
dist/meriyah.min.mjs
dist/meriyah.umd.js
dist/meriyah.umd.min.js
All transpiled with target: "ES2021"

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "unexpected": "^13.2.1"
   },
   "engines": {
-    "node": ">=10.4.0"
+    "node": ">=18.0.0"
   },
   "lint-staged": {
     "*.{ts,mts,cts,js,mjs,cjs}": "eslint --fix",
--- a/package.json
+++ b/package.json
@@ -3,17 +3,16 @@
   "version": "5.0.0",
   "description": "A 100% compliant, self-hosted javascript parser with high focus on both performance and stability",
   "main": "dist/meriyah.cjs",
-  "module": "dist/meriyah.esm.js",
-  "jsnext:main": "dist/meriyah.esm.js",
+  "module": "dist/meriyah.min.mjs",
   "browser": "dist/meriyah.umd.min.js",
   "unpkg": "dist/meriyah.umd.min.js",
-  "exports": {
-    "import": "./dist/meriyah.esm.mjs",
-    "require": "./dist/meriyah.cjs",
-    "types": "./dist/src/meriyah.d.ts"
-  },
   "types": "dist/src/meriyah.d.ts",
   "typings": "dist/src/meriyah.d.ts",
+  "exports": {
+    "types": "./dist/src/meriyah.d.ts",
+    "import": "./dist/meriyah.min.mjs",
+    "require": "./dist/meriyah.cjs"
+  },
   "license": "ISC",
   "homepage": "https://github.com/meriyah/meriyah",
   "repository": {
@@ -31,6 +30,10 @@
     {
       "name": "Chunpeng Huo",
       "url": "https://github.com/3cp"
+    },
+    {
+      "name": "Fisker Cheung",
+      "url": "https://github.com/fisker"
     }
   ],
   "keywords": [
@@ -51,7 +54,6 @@
   ],
   "files": [
     "dist",
-    "src",
     "README.md",
     "CHANGELOG.md",
     "LICENSE.md"
--- a/package.json
+++ b/package.json
@@ -30,10 +30,6 @@
     {
       "name": "Chunpeng Huo",
       "url": "https://github.com/3cp"
-    },
-    {
-      "name": "Fisker Cheung",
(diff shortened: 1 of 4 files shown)


3cp:
I locally tested it with `npm run bundle:bench` and `npm run bundle`.
I tried `npm pack` to produce a tgz file then consumed it in another project. Both CommonJS and ESM usages in Nodejs environment worked.

3cp:
Addresses #236

fisker:
How about

```diff
- dist/meriyah.cjs.js
+ dist/meriyah.cjs
- dist/meriyah.cjs.min.js
- dist/meriyah.mjs
+ dist/meriyah.min.mjs
- dist/meriyah.umd.js
+ dist/meriyah.js
- dist/meriyah.umd.min.js
+ dist/meriyah.min.js
```

The`.cjs` `.mjs` extension already indentify what they are.

The CJS version can't use on web directly, I don't think it's nessary to minify.

package.json
```diff
@@ -2,14 +2,13 @@
...
-  "jsnext:main": "dist/meriyah.esm.js",
+  "main": "dist/meriyah.cjs.js",
+  "module": "dist/meriyah.esm.mjs",
   "browser": "dist/meriyah.umd.min.js",
   "unpkg": "dist/meriyah.umd.min.js",
   "exports": {
     "import": "./dist/meriyah.esm.mjs",
-    "require": "./dist/meriyah.cjs",
+    "require": "./dist/meriyah.cjs.js",
     "types": "./dist/src/meriyah.d.ts"
```
fisker, line 12: I know current setting works , but TS  docs says

> The "types" condition should always come first in "exports".

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing

Let's move it to top.

package.json
```diff
@@ -31,6 +30,10 @@
     {
       "name": "Chunpeng Huo",
       "url": "https://github.com/3cp"
+    },
+    {
+      "name": "Fisker Cheung",
+      "url": "https://github.com/fisker"
```
fisker, line 36: I didn't contribute much to the source, I'd prefer not list myself here. I don't want see this field because huge in future.

3cp: OK :-)

3cp:
> How about
> 
> ```diff
> - dist/meriyah.cjs.js
> + dist/meriyah.cjs
> - dist/meriyah.cjs.min.js
> - dist/meriyah.mjs
> + dist/meriyah.min.mjs
> - dist/meriyah.umd.js
> + dist/meriyah.js
> - dist/meriyah.umd.min.js
> + dist/meriyah.min.js
> ```
> 
> The`.cjs` `.mjs` extension already indentify what they are.
> 
> The CJS version can't use on web directly, I don't think it's nessary to minify.

Good idea!
The `umd` might be worthy in the file name. When I wanted to manually use lib's dist file, I look those file names.

fisker:
> The umd might be worthy in the file name. When I wanted to manually use lib's dist file, I look those file names.

Okay, also prevents link change on CDN.

package.json
```diff
@@ -2,14 +2,13 @@
   "name": "meriyah",
   "version": "5.0.0",
   "description": "A 100% compliant, self-hosted javascript parser with high focus on both performance and stability",
-  "main": "dist/meriyah.cjs",
-  "module": "dist/meriyah.esm.js",
-  "jsnext:main": "dist/meriyah.esm.js",
+  "main": "dist/meriyah.cjs.js",
+  "module": "dist/meriyah.esm.mjs",
```
fisker, line 6: I think this `module` entry should link to minified version.

Import from  `https://unpkg.com/meriyah?module` will use it, jsdelivr should also use it I guess.

I think only CDNs and bundlers will look for this field.

3cp: Done.

fisker (approved):

fisker:
Good job.

fisker:
The branch is not deleted, maybe we can enable auto delete? https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-the-automatic-deletion-of-branches

fisker:
@3cp 

Do you think add `"type": "module"` will break the dist files (I guess not)?
I'm trying to convert js files (in `/scripts`) to ESM. If it's not safe we can use `.mjs` instead.

3cp:
It will affect the umd.js file.
In `/scripts`, there is another package.json file to control the local "type".

fisker:
> It will affect the umd.js file.

It's not in exports or main, so it's not possible to require or import. It can only loaded by `<script>`.

Of course ,  import it directly in our repo will throw, but this is only case I can image.

PR #286, opened 2024-07-26
fix(parser): fix annexB behaviour (webcompat) of catch param

3cp:
ECMA B.3.4 syntax error if catch param occurs in var declared names of catch block, unless catch param is binding identifier

### DIFF
--- a/scripts/run-test262.ts
+++ b/scripts/run-test262.ts
@@ -14,24 +14,7 @@ function loadList(filename: string) {
 }
 
 const unsupportedFeatures = new Set(loadList('unsupported-features.txt'));
-const whitelist = [
-  ...loadList('whitelist.txt'),
-  // Can't construct these regexps on Node.js v16
-  ...(process.version.startsWith('v16.')
-    ? [
-        'built-ins/RegExp/named-groups/non-unicode-property-names-valid.js (default)',
-        'built-ins/RegExp/named-groups/non-unicode-property-names-valid.js (strict mode)',
-        'built-ins/RegExp/property-escapes/generated/Script_-_Kawi.js (default)',
-        'built-ins/RegExp/property-escapes/generated/Script_-_Kawi.js (strict mode)',
-        'built-ins/RegExp/property-escapes/generated/Script_-_Nag_Mundari.js (default)',
-        'built-ins/RegExp/property-escapes/generated/Script_-_Nag_Mundari.js (strict mode)',
-        'built-ins/RegExp/property-escapes/generated/Script_Extensions_-_Kawi.js (default)',
-        'built-ins/RegExp/property-escapes/generated/Script_Extensions_-_Kawi.js (strict mode)',
-        'built-ins/RegExp/property-escapes/generated/Script_Extensions_-_Nag_Mundari.js (default)',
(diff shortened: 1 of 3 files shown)


scripts/run-test262.ts
```diff
@@ -14,24 +14,7 @@ function loadList(filename: string) {
 }
 
 const unsupportedFeatures = new Set(loadList('unsupported-features.txt'));
-const whitelist = [
-  ...loadList('whitelist.txt'),
-  // Can't construct these regexps on Node.js v16
-  ...(process.version.startsWith('v16.')
```
3cp, line 20: Nodejs v16 is removed from CI, and removed from package.json engine.

src/common.ts
```diff
@@ -724,12 +724,8 @@ export function addVarName(
...
       }
     }
-    if (value & (BindingKind.CatchIdentifier | BindingKind.CatchPattern)) {
-      if (
-        (value & BindingKind.CatchIdentifier) === 0 ||
-        (context & Context.OptionsWebCompat) === 0 ||
-        context & Context.Strict
-      ) {
+    if (value & BindingKind.CatchIdentifierOrPattern) {
+      if ((context & Context.OptionsWebCompat) === 0 || (value & BindingKind.CatchIdentifier) === 0) {
```
3cp, line 728: When in webcompat mode (annexB), only fail if it's CatchPattern `catch ({message}) { var message }`.

fisker, line 728: [unicorn/no-lonely-if](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-lonely-if.md)

fisker (approved):

src/common.ts
```diff
@@ -724,14 +724,11 @@ export function addVarName(
...
-        (value & BindingKind.CatchIdentifier) === 0 ||
-        (context & Context.OptionsWebCompat) === 0 ||
-        context & Context.Strict
-      ) {
-        report(parser, Errors.DuplicateBinding, name);
-      }
+    if (
+      value & BindingKind.CatchPattern ||
+      (value & BindingKind.CatchIdentifier && (context & Context.OptionsWebCompat) === 0)
+    ) {
```
3cp, line 730: The cleaned up if-logic is more readable.

fisker: Did we duplicate the `value` check?

fisker: Sorry... it's `&`..

fisker (approved):

src/common.ts
```diff
@@ -724,10 +724,11 @@ export function addVarName(
         currentScope.scopeError = recordScopeError(parser, Errors.DuplicateBinding, name);
       }
     }
-    if (value & BindingKind.CatchIdentifierOrPattern) {
-      if ((context & Context.OptionsWebCompat) === 0 || (value & BindingKind.CatchIdentifier) === 0) {
-        report(parser, Errors.DuplicateBinding, name);
-      }
+    if (
+      value & BindingKind.CatchPattern ||
+      (value & BindingKind.CatchIdentifier && (context & Context.OptionsWebCompat) === 0)
```
fisker, line 729: @3cp I didn't use bitmask much, but why are we using it for options? Shouldn't we use boolean instead?

3cp: Context uses bitmask for all options, as options can be applied together.

PR #292, opened 2024-07-27
fix(parser): async function should not skip duplication error

### LINKED ISSUE #291: Ecma spec requires separation of FunctionDeclaration and AsyncFunctionDeclaration
For the next entry in our whitelist.

`language/block-scope/syntax/redeclaration/async-function-name-redeclaration-attempt-with-async-function.js`

```
// This file was procedurally generated from the following sources:
// - src/declarations/async-function.case
// - src/declarations/redeclare/block-attempt-to-redeclare-async-function-declaration.template
/*---
description: redeclaration with AsyncFunctionDeclaration (AsyncFunctionDeclaration in BlockStatement)
esid: sec-block-static-semantics-early-errors
features: [async-functions]
flags: [generated]
negative:
  phase: parse
  type: SyntaxError
info: |
    Block : { StatementList }

    It is a Syntax Error if the LexicallyDeclaredNames of StatementList contains
    any duplicate entries.

---*/


$DONOTEVALUATE();

{ async function f() {} async function f() {} }
```

However there is an annexB (webcompat) modification, test262 also has it covered.
`annexB/language/function-code/function-redeclaration-block.js`

```
// Copyright (C) 2019 Adrian Heine. All rights reserved.
// This code is governed by the BSD license found in the LICENSE file.
/*---
description: In non-strict mode, duplicate LexicallyDeclaredNames in a block are allowed if they are bound by FunctionDeclarations
esid: sec-block-duplicates-allowed-static-semantics
es6id: B.3.3.4
flags: [noStrict]
info: |
    B.3.3.4 Changes to Block Static Semantics: Early Errors

    For web browser compatibility, that rule is modified with the addition of the **highlighted** text:

    Block: {StatementList}

    It is a Syntax Error if the LexicallyDeclaredNames of StatementList contains any duplicate entries, **unless the source code matching this production is not strict mode code and the duplicate entries are only bound by FunctionDeclarations**.
---*/

{ function a() {} function a() {} }
```

The catch is the annexB change only covered normal function, not async function. My guess is that async function is later added to the spec, so ecma doesn't want to bring the bad legacy behaviour to async function.

That sounds good, the bad part is meriyah only has one `BindingKind.FunctionLexical` which covers both normal function and async function. We would need to separate them.

3cp:
AnnexB 3.2.4 changes to Block Static Semantics: Early Errors only covers FunctionDeclarations, not AsyncFunctionDeclarations

closes #291


src/common.ts
```diff
@@ -652,8 +653,10 @@ export function addBlockName(
       scope.scopeError = recordScopeError(parser, Errors.DuplicateBinding, name);
     } else if (
       context & Context.OptionsWebCompat &&
-      value & BindingKind.FunctionLexical &&
-      origin & Origin.BlockStatement
+      (context & Context.Strict) === 0 &&
+      origin & Origin.BlockStatement &&
+      value === BindingKind.FunctionLexical &&
+      kind === BindingKind.FunctionLexical
```
3cp, line 659: > It is a Syntax Error if the [LexicallyDeclaredNames](https://262.ecma-international.org/#sec-static-semantics-lexicallydeclarednames) of [StatementList](https://262.ecma-international.org/#prod-StatementList) contains any duplicate entries, **unless the [source text matched by](https://262.ecma-international.org/#sec-algorithm-conventions-syntax-directed-operations) this production is not [strict mode code](https://262.ecma-international.org/#sec-strict-mode-code) and the duplicate entries are only bound by FunctionDeclarations.**

webcompat mode, not strict mode, in block, both duplicate entries are FunctionDeclarations

3cp: `===` makes sure they are not AsyncFunctionLexical.

src/common.ts
```diff
@@ -72,9 +72,10 @@ export const enum BindingKind {
   Const = 1 << 4,
   Class = 1 << 5,
   FunctionLexical = 1 << 6,
-  FunctionStatement = 1 << 7,
-  CatchPattern = 1 << 8,
-  CatchIdentifier = 1 << 9,
+  AsyncFunctionLexical = (1 << 7) | FunctionLexical,
```
3cp, line 75: Not sure if this is the best way to add AsyncFunctionLexical.

3cp:
Pushed another tiny unrelated fix here.

src/common.ts
fisker: I'm not familiar with the parser, but do we have the information about declared functions in scope? 

I saw there are failing tests for generator functions, will we add more like this?

3cp: Thx!
Generator function is new too. Should not carry legacy behaviour either. I will fix that.

fisker: I mean shouldn't we already have the declared functions, can't we check if function names are duplicates directly?

3cp: All normal/async/generator functions are functions. We need to check the duplicated bindings in scope are normal function, not async nor generator.

fisker: Aren't they already parsed, and we know it's an async function or not? 

Something like

```js
if (
  duplicatedFunctions.some(fn =>
    !fn.async
    && !fn.generator
    && fn.id.value === name)
  )
) {
  // ..
}
```

won't work? Or only because we avoid such test for performance?

3cp: Our lexical scope doesn't save the node (identifier, function, ...) itself, it only saves a BindingKind. There is not even start/end/position saved.

The scope is a mapping like `{ "#funcName" : BindingKind.FunctionLexical }`.

The BindingKind is the only thing we can read when checking duplication.

fisker (approved):

src/common.ts
fisker: Thanks for the explanation.

PR #293, opened 2024-07-28
feat(lexer): support regexp v flag (unicodeSets)

3cp:
Note meriyah uses JavaScrip runtime to validate regexp pattern, so the v flag only works in JS runtime that supports it. For nodejs, it's v20.0.0+.

### DIFF
--- a/scripts/run-test262.ts
+++ b/scripts/run-test262.ts
@@ -1,7 +1,10 @@
 import { parseScript, parseModule } from '../src/meriyah';
 import * as path from 'path';
 import * as fs from 'fs';
+import * as semver from 'semver';
 import run = require('test262-parser-runner');
+
+const vFlagSupported = semver.gte(process.version.slice(1), '20.0.0');
 
 function loadList(filename: string) {
   const file = path.join(__dirname, '../test/test262-parser-tests', filename);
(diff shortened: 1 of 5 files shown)


scripts/run-test262.ts
```diff
@@ -1,8 +1,11 @@
 import { parseScript, parseModule } from '../src/meriyah';
 import * as path from 'path';
 import * as fs from 'fs';
+import * as semver from 'semver';
 import run = require('test262-parser-runner');
 
+const vFlagSupported = semver.gte(process.version.slice(1), '20.0.0');
```
fisker, line 7: Why don't we use `new RegExp('', 'v')` to test?

fisker:
```suggestion
const vFlagSupported = (() => {
  try {
    return new RegExp('', 'v').flags === ‘v’
  } catch {}
})();
```

3cp: Awesome!

3cp:
What about introducing a breaking change, update package.json engine to node >=20.0.0?
We are pending major release anyway. Nodejs v18 is near EOL.

fisker (approved):

fisker:
> What about introducing a breaking change, update package.json engine to node >=20.0.0? We are pending major release anyway. Nodejs v18 is near EOL.

I don't know, but the `RegexLiteral.value` is fine to have a `null` value.

PR #295, opened 2024-07-31
fix(parser): only class static non-private property cannot be named 'prototype'

3cp:


fisker (approved):
What a surprise.

3cp:
Added another small patch here for future reserved keywords.
I verified those modified unit tests with nodejs runtime.

fisker (approved):

PR #298, opened 2024-08-02
feat: support import attributes with "assert" keyword

zermelo-wisen:
This PR adds support for the old `assert` keyword syntax (import-assertions). The resulting node is the same as with the `with` syntax, i.e., `ImportAttributes` are added to the `attributes` member instead of the `assertions` member.

```typescript
interface ImportDeclaration {
    attributes: [ ImportAttribute ];
}
```

References:
- [Import Attributes](https://github.com/estree/estree/blob/master/stage3/import-attributes.md)
- [Import Assertions](https://github.com/estree/estree/blob/master/stage3/import-assertions.md)

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2536,7 +2536,10 @@ function parseImportDeclaration(
   };
 
   if (context & Context.OptionsNext) {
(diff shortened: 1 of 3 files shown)


fisker:
It's deprecated, I don't think we should support.

3cp:
I read it's deprecated and replaced by import-attributes.

PR #299, opened 2024-08-02
fix(parser): escaped reserved keyword can be used as class element name

3cp:
Just learnt class property name and method name are NOT Identifier in ECMA spec, it's called ClassElementName. Similar to the PropertyName in ObjectLiteral, which is NOT Identifier either.

Keywords cannot be Identifier, but can be ClassElementName and PropertyName.

```js
class A {
 let = 1;
 function = 2;
 static voi\u0064 = 3; // escaped void
 async *var() {}
}
```
The above are all valid. :)


fisker (approved):
👍

PR #301, opened 2024-08-03
fix(parser): dot property name can be escaped reserved or future reserved

3cp:


3cp:
Second commit removed duplicated code, also added private identifier support in optional chaining.

src/parser.ts
```diff
@@ -4149,7 +4148,12 @@ export function parseOptionalChain(
  * @param context Context masks
  */
 export function parsePropertyOrPrivatePropertyName(parser: ParserState, context: Context): any {
-  if ((parser.getToken() & (Token.IsIdentifier | Token.Keyword)) === 0 && parser.getToken() !== Token.PrivateField) {
+  if (
+    (parser.getToken() & (Token.IsIdentifier | Token.Keyword)) === 0 &&
+    parser.getToken() !== Token.EscapedReserved &&
+    parser.getToken() !== Token.EscapedFutureReserved &&
```
fisker, line 4154: Are you sure this is needed? Shouldn't the property name already allow reserved since ES5?

fisker: Ah, the token is neither `Identifier` nor `Keyword`?

fisker (approved):

src/parser.ts
3cp: The normal usage was covered `x.void`, but not the escaped form. Our `EscapedReserved` and `EscapedFutureReserved` didn't carry any shared mask bit.

We should have a new mask bit called `IsIdentifierName` I think. In ECMA spec, identifier name can be anything, then identifer removed reserved words. PropertyName and ClassElementName uses IdenifierName, not Identifier.

3cp: However, ESTree uses type:identifier for all cases, bit confusing.

PR #302, opened 2024-08-05
fix(parser): fix simple params list and strict reserved check on func

### LINKED ISSUE #296: How to handle future reserved keyword in function name/param without backtrace
Future reserved keywords can be used as identifier, but ONLY in non-strict mode.

Valid code:
```js
function a(let) {}
```

In valid code:
```js
"use strict"; function a(let) {}
```

Meriyah can handle the above two cases.
BTW, the second commit in #295 handled escaped identifier like `l\u0065t` (let).

However, missed following tricky one:

```js
function a(let) { "use strict"; }
```
This line should fail with "Unexpected strict mode reserved word" at "let".

The strict mode turned on by "use strict" needs to be applied to function name and parameters too. Since meriyah doesn't do back-trace, this is very tricky to implement.

3cp:
Has to revert some changes for previous #295. Now it handles escaped reserved words in non-strict mode properly.

closes #296


3cp:
Unfortunately has to add `language/statements/let/syntax/escaped-let.js` to whitelist.
But it's not a introduced bug, but exposing existing bug.
It's about `l\u0065t` cannot be used for in "let" declaration.

3cp:
I will settle with the added `language/statements/for-await-of/head-lhs-async.js` for now.

```js
let async;

async function fn() {
  for await (async of [7]);
}
```

Now meriyah thinks "of" after "async" is an identifier, expecting `async of => ...`.
Have to make sure "of" after "async" is the "of" keyword in the for statement.

3cp:
Who added support of unicode escaping in identifier to ecma spec, the fellow is plain crazy...
Just use an editor that is capable to display native unicode character. Nobody writes JS on a mainframe terminal.
Totally a feature nobody needs.

3cp:
I think I can fix that `async of`.

3cp:
ecma 14.7.5

for ( [lookahead ∉ { let, async of }] [LeftHandSideExpression](https://tc39.es/ecma262/#prod-LeftHandSideExpression)[?Yield, ?Await] of 
  [AssignmentExpression](https://tc39.es/ecma262/#prod-AssignmentExpression)[+In, ?Yield, ?Await] ) [Statement](https://tc39.es/ecma262/#prod-Statement)[?Yield, ?Await, ?Return]

fisker (approved):

PR #307, opened 2024-08-06
fix(parser): class static block disallows "await" as identifier

3cp:
ClassStaticBlockStatementList :
  StatementList[~Yield, +Await, ~Return]opt

static block is InAwaitContext.


3cp:
ECMA has Grammatical Parameters `[In, Yield, Await, Return, Tagged]`.
We have `Context.DisallowIn` (for In ?), `Context.InYieldContext` (confirmed for Yield), `Context.InAwaitContext` (confirmed for Await), `Context.TaggedTemplate` (for Tagged ?).

It looks like we have missing `Context.InReturnContext` to control the behaviour of ReturnStatement. There are missing test262 cases in whitelist.

fisker (approved):

3cp:
Thx as always!

PR #308, opened 2024-08-07
fix(parser): class static block has no return statement

3cp:
Context.InReturnContext is added to match the spec, replaced Context.OptionsGlobalReturn.


3cp:
Added another tiny fix.

3cp:
Few more tiny fixes.

fisker (approved):

PR #309, opened 2024-08-07
fix(parser): "use strict" directive could be after invalid string

### LINKED ISSUE #149: Use Strict Directive needs to be preflighted
Found this issue while cleaning up test262 file, not sure if this is possible in meriyah.
It's not a serious bug in real world, as people always put use strict directive at beginning of a scope.

Meriyah rejects following code as per spec:
```
> parse('"use strict"; "\\1";')
Uncaught:
[ParseError [SyntaxError]: [1:16]: Octal escape sequences are not allowed in strict mode
] {
  index: 16,
  line: 1,
  column: 16,
  description: '[1:16]: Octal escape sequences are not allowed in strict mode',
  loc: { line: 1, column: 16 }
}
```

But failed to do so if the use strict directive is after the problematic code.
```
> parse('"\\1"; "use strict";')
{
  type: 'Program',
  sourceType: 'script',
  body: [
    { type: 'ExpressionStatement', expression: [Object] },
    { type: 'ExpressionStatement', expression: [Object] }
  ]
}
```

3cp:
closes #149


kenbern:
@3cp what is I believe is one of the author of this code - KFlash - has an almost identical code on his repo. Seafox. Everything fixed there

fisker (approved):

3cp:
Thx!

PR #315, opened 2024-08-14
fix(parser): fix jsx with comments in tag

### LINKED ISSUE #185: JSX with comments not correctly parsed
Examples:

```jsx
<a><// line
/a>;
```

```jsx
<a></* block */
/a>;
```

```jsx
</* open fragment */>
</ /* close fragment */>;
```

3cp:
closes #185

### DIFF
--- a/src/lexer/jsx.ts
+++ b/src/lexer/jsx.ts
@@ -68,42 +68,43 @@ export function scanJSXToken(parser: ParserState, context: Context): Token {
       if (parser.getToken() !== Token.LessThan)
         report(parser, Errors.UnexpectedToken, KeywordDescTable[parser.getToken() & Token.Type]);
 
-      // Possible white space or comments between "<" and "/".
+      // Possible whitespace or comments between "<" and "/".
       const {
         index,
         line,
         column,
         tokenIndex,
+        tokenColumn,
+        tokenLine,
         startIndex,
         startColumn,
         startLine,
-        tokenColumn,
-        tokenLine,
         currentChar,
         tokenValue,
         tokenRaw,
         tokenRegExp
       } = parser;
       const tokenAfter = scanSingleToken(parser, context, LexerState.None);
+      parser.tokenIndex = tokenIndex;
+      parser.tokenColumn = tokenColumn;
+      parser.tokenLine = tokenLine;
+      parser.startIndex = startIndex;
+      parser.startColumn = startColumn;
+      parser.startLine = startLine;
+      // Punctuator doesn't change tokenValue or raw, just keep previous
+      parser.tokenValue = tokenValue;
+      parser.tokenRaw = tokenRaw;
+      parser.tokenRegExp = tokenRegExp;
 
       if (tokenAfter === Token.Divide) {
         // Rewrite LessThan + Divide as JSXClose
         parser.setToken(Token.JSXClose, true);
       } else {
-        // Reset if not merged
+        // Restore if not merged
         parser.index = index;
         parser.line = line;
         parser.column = column;
-        parser.tokenIndex = tokenIndex;
-        parser.startIndex = startIndex;
-        parser.startColumn = startColumn;
-        parser.startLine = startLine;
-        parser.tokenColumn = tokenColumn;
-        parser.tokenLine = tokenLine;
         parser.currentChar = currentChar;
-        parser.tokenValue = tokenValue;
-        parser.tokenRaw = tokenRaw;
-        parser.tokenRegExp = tokenRegExp;
       }
 
       break;
--- a/src/lexer/jsx.ts
+++ b/src/lexer/jsx.ts
@@ -13,9 +13,9 @@ import { decodeHTMLStrict } from './decodeHTML';
  */
 export function scanJSXAttributeValue(parser: ParserState, context: Context): Token {
   // skip "=" before the value
-  parser.startPos = parser.tokenPos = parser.index;
-  parser.startColumn = parser.colPos = parser.column;
-  parser.startLine = parser.linePos = parser.line;
+  parser.startIndex = parser.tokenIndex = parser.index;
+  parser.startColumn = parser.tokenColumn = parser.column;
+  parser.startLine = parser.tokenLine = parser.line;
   parser.setToken(
     CharTypes[parser.currentChar] & CharFlags.StringLiteral
       ? scanJSXString(parser, context)
@@ -42,7 +42,7 @@ export function scanJSXString(parser: ParserState, context: Context): Token {
   if (char !== quote) report(parser, Errors.UnterminatedString);
   parser.tokenValue = parser.source.slice(start, parser.index);
   advanceChar(parser); // skip the quote
-  if (context & Context.OptionsRaw) parser.tokenRaw = parser.source.slice(parser.tokenPos, parser.index);
+  if (context & Context.OptionsRaw) parser.tokenRaw = parser.source.slice(parser.tokenIndex, parser.index);
   return Token.StringLiteral;
 }
 
@@ -52,9 +52,9 @@ export function scanJSXString(parser: ParserState, context: Context): Token {
  * @param parser The parser object
  */
 export function scanJSXToken(parser: ParserState, context: Context): Token {
-  parser.startPos = parser.tokenPos = parser.index;
-  parser.startColumn = parser.colPos = parser.column;
-  parser.startLine = parser.linePos = parser.line;
+  parser.startIndex = parser.tokenIndex = parser.index;
+  parser.startColumn = parser.tokenColumn = parser.column;
+  parser.startLine = parser.tokenLine = parser.line;
 
   if (parser.index >= parser.end) return parser.setToken(Token.EOF);
 
@@ -73,12 +73,12 @@ export function scanJSXToken(parser: ParserState, context: Context): Token {
         index,
         line,
         column,
-        tokenPos,
-        startPos,
+        tokenIndex,
+        startIndex,
         startColumn,
         startLine,
-        colPos,
-        linePos,
+        tokenColumn,
+        tokenLine,
         currentChar,
         tokenValue,
         tokenRaw,
@@ -94,12 +94,12 @@ export function scanJSXToken(parser: ParserState, context: Context): Token {
         parser.index = index;
         parser.line = line;
         parser.column = column;
-        parser.tokenPos = tokenPos;
-        parser.startPos = startPos;
+        parser.tokenIndex = tokenIndex;
+        parser.startIndex = startIndex;
         parser.startColumn = startColumn;
         parser.startLine = startLine;
-        parser.colPos = colPos;
-        parser.linePos = linePos;
+        parser.tokenColumn = tokenColumn;
+        parser.tokenLine = tokenLine;
         parser.currentChar = currentChar;
         parser.tokenValue = tokenValue;
         parser.tokenRaw = tokenRaw;
@@ -133,7 +133,7 @@ export function scanJSXToken(parser: ParserState, context: Context): Token {
         if (CharTypes[parser.currentChar] & CharFlags.JSXToken) break;
       }
 
-      const raw = parser.source.slice(parser.tokenPos, parser.index);
+      const raw = parser.source.slice(parser.tokenIndex, parser.index);
       if (context & Context.OptionsRaw) parser.tokenRaw = raw;
       parser.tokenValue = decodeHTMLStrict(raw);
       parser.setToken(Token.JSXText);
--- a/src/lexer/jsx.ts
+++ b/src/lexer/jsx.ts
@@ -1,9 +1,8 @@
 import { CharFlags, CharTypes } from './charClassifier';
-import { Chars } from '../chars';
-import { Token } from '../token';
+import { KeywordDescTable, Token } from '../token';
 import { ParserState, Context } from '../common';
 import { report, Errors } from '../errors';
-import { advanceChar, LexerState, TokenLookup, scanSingleToken, scanNewLine, consumeLineFeed } from './';
+import { advanceChar, LexerState, TokenLookup, scanSingleToken, scanNewLine, consumeLineFeed, nextToken } from './';
 import { decodeHTMLStrict } from './decodeHTML';
 
 /**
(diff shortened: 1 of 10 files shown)


src/lexer/jsx.ts
```diff
@@ -64,12 +63,46 @@ export function scanJSXToken(parser: ParserState, context: Context): Token {
...
+        parser.startPos = startPos;
+        parser.startColumn = startColumn;
+        parser.startLine = startLine;
+        parser.colPos = colPos;
+        parser.linePos = linePos;
+        parser.currentChar = currentChar;
+        parser.tokenValue = tokenValue;
+        parser.tokenRaw = tokenRaw;
+        parser.tokenRegExp = tokenRegExp;
       }
```
3cp, line 106: This is the only way I could come out. Reuse the standard scanner so white spaces and comments will be handled.

Token "<" and "/" need to be merged into "</". Otherwise just return "<" (done by first nextToken() call), and revert effect of the extra scanSingleToken.
In short, we peek ahead and restore if needed.

src/lexer/scan.ts
```diff
@@ -304,7 +304,7 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
               // Check that it's not a comment start.
               if (index < parser.end) {
                 ch = source.charCodeAt(index);
-                if (ch === Chars.Asterisk || ch === Chars.Slash) break;
+                if (ch === Chars.Asterisk || ch === Chars.Slash) return Token.LessThan;
```
3cp, line 307: An existing bug.

src/parser.ts
```diff
@@ -9102,7 +9102,7 @@ export function parseJSXClosingFragment(
   if (inJSXChild) {
     consume(parser, context, Token.GreaterThan);
   } else {
-    consume(parser, context, Token.GreaterThan);
+    scanJSXToken(parser, context);
   }
```
3cp, line 9106: Another existing bug due to copy paste I assume. Note the previous two branches are same.

3cp:
Our jsx support is dirty as hell.
I would love to crop it from meriyah :-)

We need a better test suite for jsx, I am not confident at all about my changes.

src/lexer/jsx.ts
fisker: Can we save these to a object (Object.pick())? and restore with `Object.assign()` later? too many to me.

Or add `save()` `restore()`  methods to Parse, like html's ContentRendering2D

fisker: Maybe we can use `using`? I heard that TS already support.

3cp: Oh, fancy latest resource management :-)
I will sit with the PR for few days until we are satisfied.

kenbern: I played with jsx yesterday while I ported it to typescript from my assembly code. 

Tricky things to remember is that jsx has it's own special comments, jsx comments and some other nasty features that no one implement. 

You also need to preserve collapsed jsx whitespace.  

The >> is an edge case that is poorly handled by meriyah. Running the normal scanner for jsx will give issue with some punctuators. 

Use a separate scanner

3cp: Maybe two new APIs to ParserState, like `parser.snapshot(): ParserStateSnapshot` and `parser.restore(snapshot: ParserStateSnapshot)`.
Or `parser.createCheckPoint(): void` and `parser.restoreCheckPoint(): void`?

kenbern: What is snapshot?

kenbern: I got inspiration from meriyah lexer in my rewrite - build the same way  -  but I changed the approach a little so you can check ahead on next token without any backtracking and perf loss. That way you could check if token + 1 is > for >> edge cases. 

Or simply refactor tokens and return a token.jsx node from lexer itself. Similar approach I did for solving ?. edge case

kenbern: You can  change meriyah to return token from sst like this and then handle ast nodes like arrays instead of object and you can do x[1] to get what you want. 

Example

tokens.push({
      type: TokenType.NUMBER,
      value: result,
      loc: ((state.tokenStartLine << 20) | (state.tokenStartColumn << 10) | ((state.lineNumber << 20) | state.column)) >>> 0,
      flags: state.tokenFlags,
    });

fisker: > Maybe two new APIs to ParserState, like `parser.snapshot(): ParserStateSnapshot` and `parser.restore(snapshot: ParserStateSnapshot)`. Or `parser.createCheckPoint(): void` and `parser.restoreCheckPoint(): void`?

This sounds good

3cp: To continue on updated code.

src/lexer/jsx.ts
```diff
@@ -64,12 +63,48 @@ export function scanJSXToken(parser: ParserState, context: Context): Token {
...
+      if (tokenAfter === Token.Divide) {
+        // Rewrite LessThan + Divide as JSXClose
+        parser.setToken(Token.JSXClose, true);
       } else {
-        parser.setToken(Token.LessThan);
+        // Restore if not merged
+        parser.index = index;
+        parser.line = line;
+        parser.column = column;
+        parser.currentChar = currentChar;
```
3cp, line 107: It's not a simple restore if not merged. Even for merged, has to take care of token start position, otherwise onToken will miss all chars before "/".

The other choice is to keep "< /" as two tokens "<" and "/", that's natural, but require some refactoring of jsx parsing. Might be better.

src/lexer/jsx.ts
kenbern: @3cp if you can wait a week or something then I can show you an example on correctly implementation. 

I'm busy as of now fine in tuning my  type system for my parser - ai insperated type system so no need for TS. 

BTW! JSX has it's own annotations and comments. The comment call back in meriya should be adjusted to handle them. 

And some of this comments obviously gives instruction on how to add additiiycide

kenbern:
@3cp have fun with this tests! 


```ts


function MyComponent({ isLoading }) {
  const [data, setData] = useState(null);

  useEffect(() => {
    if (!isLoading) {
      fetchSomething()
        .then(setData)
        .catch((error) => console.error(error));
    }
  }, [isLoading]);

  return (
    <div>
      {isLoading ? <p>Loading...</p> : <p>{data}</p>}
    </div>
  );
}


function MyComponent({ isLoading }) {
  // Assume 'fetchSomething' is an asynchronous function
  const data = isLoading ? null : fetchSomething();

  return (
    <div>
      {isLoading ? <p>Loading...</p> : <p>{data}</p>}
    </div>
  );
}


function MyComponent({ items }) {
  const handleItemChange = (event) => {
    // Get the changed item based on event.target 
    // ... logic to update state and re-render based on the specific item
  };

  return (
    <ul>
      {items.map((item) => (
        <li key={item.id}>
          <input type="checkbox" onChange={handleItemChange} /> {/* Each checkbox shares the same handler */}
          {item.name}
        </li>
      ))}
    </ul>
  );
}


function MyComponent({ isActive }) {
  const style = {
    color: isActive ? 'green' : 'red',
    fontSize: isActive ? '20px' : '16px',
    // ... more styles
  };

  return <div style={style}>Hello!</div>;
}




function MyComponent({ data }) {
  const computedValue = data.map((item) => {
    // Complex logic to compute something from item
  });

  return (
    <div>
      {computedValue.map((value) => (
        <span key={value.id}>{value}</span>
      ))}
    </div>
  );
}


function Tree({ data }) {
  return (
    <ul>
      {data.map((item) => (
        <li key={item.id}>
          {item.name}
          {item.children && <Tree data={item.children} />}
        </li>
      ))}
    </ul>
  );
}


function Counter({ initialCount }) {
  const [count, setCount] = useState(initialCount);

  const handleClick = () => {
    setCount(count + 1);
    // Potential issue if this action depends on the updated count
    doSomethingElse(count);
  };

  return <button onClick={handleClick}>Increment</button>;
}


function MyComponent({ items }) {
  return (
    <ul>
      {items.map((item) => (
        <li key={item.id}> {/* Potential key issue */}
          {item.active ? <span>{item.name}</span> : null}
        </li>
      ))}
    </ul>
  );
}




function MyComponent({ items }) {
  return (
    <ul>
      {items.map((item, index) => (
        <li key={`${item.id}-${index}`}> 
          <>{item.active && <span>{item.name}</span>}</>
        </li>
      ))}
    </ul>
  );
}


function Counter({ initialCount }) {
  const [count, setCount] = useState(initialCount);

  const handleClick = () => {
    setCount(count + 1);
    // Potential issue if this action depends on the updated count
    doSomethingElse(count);
  };

  return <button onClick={handleClick}>Increment</button>;
}


const handleClick = () => {
  setCount((prevCount) => {
    const newCount = prevCount + 1;
    doSomethingElse(newCount); // Now uses the updated count
    return newCount;
  });
};




function Tree({ data }) {
  return (
    <ul>
      {data.map((item) => (
        <li key={item.id}>
          {item.name}
          {item.children && <Tree data={item.children} />}
        </li>
      ))}
    </ul>
  );
}


function MyComponent({ text }) {
  const regex = /<[^>]+>/g; // Matches any HTML tag
  const sanitizedText = text.replace(regex, ''); // Potentially dangerous

  return <p>{sanitizedText}</p>;
}



function MyComponent({ data }) {
  const computedValue = data.map((item) => {
    // Complex logic to compute something from item
  });

  return (
    <div>
      {computedValue.map((value) => (
        <span key={value.id}>{value}</span>
      ))}
    </div>
  );
}
```

3cp:
> @3cp have fun with this tests!

Passed with no issue. Your example actually is well formatted code, no edge cases there.

kenbern:
Try this

```ts
import React, { useState } from 'react';

const ProductCard = ({ product }) => {
  const [isSelected, setIsSelected] = useState(false);

  const cardStyles = {
    border: `2px solid ${isSelected ? 'blue' : 'gray'}`,
    padding: '20px',
    borderRadius: '8px',
    width: '300px',
    margin: '10px',
    transition: 'transform 0.2s ease',
    transform: isSelected ? 'scale(1.05)' : 'scale(1)', 
    '&:hover': {
      boxShadow: '0px 4px 8px rgba(0, 0, 0, 0.2)', 
    },
    '& .product-title': { // Nested selector 
      color: product.available ? 'green' : 'red',
    } 
  };

  return (
    <div 
      className="product-card"
      style={cardStyles} 
      onClick={() => setIsSelected(!isSelected)} 
    >
      <h2 className="product-title">{product.name}</h2>
      <p>{product.description}</p>
      <p>Price: ${product.price}</p>
      {!product.available && <span className="out-of-stock">Out of Stock</span>}
    </div>
  );
};

export default ProductCard;



import React from 'react';

const MyComponent = () => {
  const message = "This could be tricky!";

  return (
    <div>
      {/* 
        A multi-line comment in JSX 
        Notice it needs to be within curly braces {/* */}
      */}

      <h1>{message}</h1> {/* Inline comment within JSX */}

      {/* 
        Edge Case: Commenting out JSX with JavaScript expressions inside
        {/* <div>{message}</div> */} 
        This will cause an error because the inner JSX expression is still being evaluated
      */}

      <p>
        {/* 
          Another edge case: JSX comment inside a string 
          This will be rendered as a string, not a comment: {/* Not a comment */}
        </p> 

      {/* 
        Ambiguity with JavaScript:
        JSX uses curly braces {} for embedding expressions, 
        which can sometimes look confusing within JavaScript code:

        const someValue = {x: 10}; // This is a JavaScript object, NOT JSX
        <div style={{ color: 'red' }}>  // JSX using an object for styles
          {someValue.x} {/* Accessing JavaScript variable */}
        </div> 
      */}
      
      {/* 
         Handling self-closing tags in JSX:
         - In HTML, some tags are self-closing (e.g., <br />, <img src="..."/>)
         - In JSX, you can use either syntax, but for consistency, 
           it's generally recommended to use the self-closing syntax with a slash:
      */}
      <br />
      <hr />
      <img src="/logo.png" alt="Logo" /> 

    </div>
  );
};

export default MyComponent;



import React from 'react';

const MyComponent = ({ title, ...otherProps }) => {
  const buttonStyles = {
    backgroundColor: 'blue',
    color: 'white',
    padding: '10px',
  };

  return (
    <div>
      <h1>{title}</h1>
      <button 
        style={buttonStyles}
        {...otherProps} 
        onClick={() => console.log('Clicked!')}
      >
        Click Me
      </button>
    </div>
  );
};

// Usage: 
<MyComponent title="Welcome" disabled className="custom-button" />

import React from 'react';

const Heading = ({ level, children }) => {
  const TagName = `h${level}`; // Dynamically determine heading level

  return <TagName>{children}</TagName>; 
};

const MyComponent = () => {
  const showHeading = true;
  const headingLevel = 3; 

  return (
    <div>
      {showHeading && <Heading level={headingLevel}>Dynamic Heading</Heading>}
      {/* Conditional rendering of the entire Heading component */}
    </div>
  );
};

export default MyComponent;


import React from 'react';

const MyComponent = () => {
  return (
    <div>
      <h1>Main Heading</h1>
      {/* Use a React Fragment to avoid adding an extra div */}
      <React.Fragment>
        <p>Paragraph 1</p>
        <p>Paragraph 2</p>
      </React.Fragment>
    </div>
  );
};

export default MyComponent;











```

3cp:
Your code is wrong. Eslint, tsc all report same error.

```
      {/* 
        A multi-line comment in JSX 
        Notice it needs to be within curly braces {/* */}
      */} <-- here is wrong
```

Even syntax highlight in github here shows that's wrong. If your parser thinks that's correct, that's your bug.

Please don't flood your code here. If you find a meriyah bug, raise an issue with a reproducible code sample, thx!

kenbern:
The test I provided comes from jsx and react. Flooded? I gave you tests from react library invented of jsx. Nothing to do with private vode.

3cp:
The code is wrong as I highlighted. That's why pushed back.

kenbern:
Yeah someone got it wrong

https://www.geeksforgeeks.org/how-to-comment-in-jsx/

fisker (approved):

3cp:
@fisker I will hold this one. Doing another experiment with separate "<" and "/".

kenbern:
@3cp look at this gist and you get correct code for handling jsx comments and parse jsx. 

Easy to adapt to meriyah

https://gist.github.com/kenbern/41943b9b9d2524dfeac00d68ab8fe644

3cp:
I got the other way working, but I will merge this one first, because it will make diff for next one easier to read.

PR #317, opened 2024-08-16
fix(parser): fix jsx parsing on '< / >'

### LINKED ISSUE #185: JSX with comments not correctly parsed

3cp:
closes #185

This is better I think. Previous fix will take "< /* xx */ /" as one token, that doesn't make sense, now it's "<" and "/" two tokens.

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -393,7 +393,7 @@ export function validateBindingIdentifier(
   // The BoundNames of LexicalDeclaration and ForDeclaration must not
   // contain 'let'. (CatchParameter is the only lexical binding form
   // without this restriction.)
-  if (kind & (BindingKind.Let | BindingKind.Const) && (t & Token.LetKeyword) === Token.LetKeyword) {
+  if (kind & (BindingKind.Let | BindingKind.Const) && (t & Token.Type) === (Token.LetKeyword & Token.Type)) {
     report(parser, Errors.InvalidLetConstBinding);
   }
 
--- a/src/lexer/jsx.ts
+++ b/src/lexer/jsx.ts
@@ -1,9 +1,10 @@
 import { CharFlags, CharTypes } from './charClassifier';
-import { KeywordDescTable, Token } from '../token';
+import { Token } from '../token';
 import { ParserState, Context } from '../common';
 import { report, Errors } from '../errors';
-import { advanceChar, LexerState, TokenLookup, scanSingleToken, scanNewLine, consumeLineFeed, nextToken } from './';
+import { advanceChar, LexerState, scanSingleToken, scanNewLine, consumeLineFeed } from './';
 import { decodeHTMLStrict } from './decodeHTML';
+import { Chars } from '../chars';
 
 /**
  * Scans JSX attribute value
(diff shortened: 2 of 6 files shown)


src/common.ts
```diff
@@ -393,7 +393,7 @@ export function validateBindingIdentifier(
   // The BoundNames of LexicalDeclaration and ForDeclaration must not
   // contain 'let'. (CatchParameter is the only lexical binding form
   // without this restriction.)
-  if (kind & (BindingKind.Let | BindingKind.Const) && (t & Token.LetKeyword) === Token.LetKeyword) {
+  if (kind & (BindingKind.Let | BindingKind.Const) && (t & Token.Type) === (Token.LetKeyword & Token.Type)) {
```
3cp, line 396: It was a bug made it very sensitive with the value behind Token.LetKeyword. There are two more same fixes in parser.ts.

Note tsc will replace `(Token.LetKeyword & Token.Type)` with single number, so that should not be an issue.

src/lexer/jsx.ts
```diff
@@ -47,109 +48,65 @@ export function scanJSXString(parser: ParserState, context: Context): Token {
 }
 
 /**
- * Scans JSX token
+ * consume Token.LessThan, Token.LeftBrace, or Token.JSXText
```
3cp, line 51: Only handle 3 cases, nothing else.

src/lexer/scan.ts
```diff
@@ -297,18 +297,6 @@ export function scanSingleToken(parser: ParserState, context: Context, state: Le
...
-              const index = parser.index + 1;
-
-              // Check that it's not a comment start.
-              if (index < parser.end) {
-                ch = source.charCodeAt(index);
-                if (ch === Chars.Asterisk || ch === Chars.Slash) return Token.LessThan;
-              }
-              advanceChar(parser);
-              return Token.JSXClose;
-            }
```
3cp, line 311: No more JSXClose.

src/token.ts
```diff
@@ -62,142 +62,140 @@ export const enum Token {
   QuestionMark = 22, // ?
   SingleQuote  = 23, // '
   DoubleQuote  = 24, // "
-  JSXClose     = 25, // </
-  JSXAutoClose = 26, // />
```
3cp, line 65: Removed two tokens def.

test/parser/miscellaneous/onToken.ts
```diff
@@ -220,9 +224,35 @@ describe('Miscellaneous - onToken', () => {
       { token: 'Punctuator', start: 0, end: 1, value: '<' },
       { token: 'Identifier', start: 2, end: 3, value: 'a' },
       { token: 'Punctuator', start: 3, end: 4, value: '>' },
-      { token: 'Punctuator', start: 4, end: 7, value: '< /' },
+      { token: 'Punctuator', start: 4, end: 5, value: '<' },
+      { token: 'Punctuator', start: 6, end: 7, value: '/' },
```
3cp, line 228: The separate tokens make more sense.

fisker (approved):
Looks like a lot of work, thank you!

PR #318, opened 2024-08-17
feat(parser): add token start and end index/line/column to ParseError

### LINKED ISSUE #156: ParserError is missing end position
I need the end position of any parser error for my linter, but this is not an available property in the ParserError class.

3cp:
Meriyah actually reported the end position in ParseError, this change added tokenIndex/tokenLine/tokenColumn to ParseError as the start position, together with end position in index/line/column. User can catch the error from meriyah and then uses tool like @babel/code-frame to visualise the failing token.

closes #156

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -202,6 +202,9 @@ export interface ScopeError {
   index: number;
   line: number;
   column: number;
+  tokenIndex: number;
+  tokenLine: number;
+  tokenColumn: number;
 }
 
 /**
@@ -579,13 +582,16 @@ export function createArrowHeadParsingScope(parser: ParserState, context: Contex
  * @param type Errors type
  */
 export function recordScopeError(parser: ParserState, type: Errors, ...params: string[]): ScopeError {
-  const { index, line, column } = parser;
+  const { index, line, column, tokenIndex, tokenLine, tokenColumn } = parser;
   return {
     type,
     params,
     index,
     line,
-    column
+    column,
+    tokenIndex,
+    tokenLine,
+    tokenColumn
   };
 }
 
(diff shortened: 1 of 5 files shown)


fisker:
Are you saying the error range should be "tokenIndex"to "index"?

I don't know, this seems confusing me. I didn't see other parser use those names either. Why not `.start`, `.end`?

3cp:
I agree, the naming needs cleanup.

I didn't want to introduce breaking change. It can be changed relatively easily.

Previously, we have `index`, `line`, `column` and `loc: { line, column }`. The `loc: {line, column }` is said to match acorn API.

For reference, there was a previous attempt #157 that changes to `loc: { start: { line, column }, end: { line, column } }`.

3cp:
I will have a look what other parsers report.

3cp:
Every parser reported in different way. Let's forget about acorn API shape.

I introduced a breaking change. Changed ParseError to expose

```
start: number
end: number;
range: [ number, number ]
loc: {
  start: { line: number, column: number },
  end: { line: number, column: number }
}
```

Just like what we have in AST node. This should be quite clear.

fisker:
How about introduce new loc just for errors

```
loc: {
  line: number, // Maybe non-enumerable getter?
  column: number,
  start: { line: number, column: number },
  end: { line: number, column: number }
}

3cp:
We don't have to match acorn api shape to have loc.line and loc.column. Breaking change I think it's acceptable.

Parsers do very differently to each other. For example, typescript only provides start and length without line/column.

Let's see what's natural for us.

fisker (approved):

PR #319, opened 2024-08-20
two fixes around template string

3cp:


3cp:
Add another easy fix for "yield\n*1".

fisker (approved):
I didn't know the cooked value can be null. TIL

3cp:
ESTree spec says cooked value must be null for tagged template with invalid escape.

PR #323, opened 2024-08-23
feat(parser): implement lexical analysis of private identifiers

### LINKED ISSUE #322: Scope analysis of private identifier
ECMA spec requires static analysis of private identifier.

For example

```
unknown: Private name #y is not defined. (2:19)

  1 | class A {
> 2 |   fn() { something.#y }
    |                    ^
  3 |   #x
  4 | }
```

If line2 is `something.#x`, it is valid, but `something.#y` is invalid because `#y` is not defined in the enclosed class definition.

Since the definition `#x` or `#y` could be after the reference `something.#y`, the parser needs to hold the references and do check when class definition/expression block ends.

3cp:
closes #322


3cp:
Whitelist finally goes down to sub-100 :-)

src/common.ts
```diff
@@ -757,6 +781,94 @@ export function addVarName(
...
+
+/**
+ * Adds a private identifier reference
+ *
+ * @param parser Parser state
+ * @param scope PrivateScopeState
+ * @param name Binding name
+ */
+export function addPrivateIdentifierRef(parser: ParserState, scope: PrivateScopeState, name: string): void {
+  if (!(name in scope.refs)) scope.refs[name] = [];
```
fisker, line 828:
```suggestion
  scope.refs[name] ??= [];
```

fisker (approved):

PR #324, opened 2024-08-23
fix(parser): class is implicit strict mode, but decorator before class is not

3cp:


fisker (approved):
What a strange world.

3cp:
Decorator spec is not in ecma spec yet. So it still could change :-)

PR #325, opened 2024-08-23
chore: update dependencies

fisker:


scripts/run-test262.ts
```diff
@@ -40,12 +40,17 @@ type Test = {
...
+    return true;
+  }
+
   const features = test.attrs.features ?? [];
   return features.some((feature) => unsupportedFeatures.has(feature));
 }
 
 async function runTest() {
   await run(parse, {
+    testsDirectory: path.dirname(require.resolve('test262/package.json')),
```
fisker, line 53: On my machine, the default value uses a different copy from 'test262-parser-runner', adding this will enforce to use the correct version in package.json.

3cp (approved):

3cp:
Thx!

PR #328, opened 2024-08-24
fix(parser): in strict mode, escaped future reserved can be used as identifier but not destructible

3cp:


src/parser.ts
```diff
@@ -7015,7 +7015,13 @@ export function parseObjectLiteralOrPattern(
       let state = PropertyKind.None;
       let key: ESTree.Expression | null = null;
       let value;
-      if (parser.getToken() & Token.IsIdentifier || parser.getToken() === Token.EscapedReserved) {
+      if (
+        parser.getToken() & Token.IsIdentifier ||
+        parser.getToken() === Token.EscapedReserved ||
+        parser.getToken() === Token.EscapedFutureReserved
+      ) {
+        if (parser.getToken() === Token.EscapedFutureReserved) destructible |= DestructuringKind.CannotDestruct;
```
3cp, line 7023: Only in strict mode, our scanner emits Token.EscapedFutureReserved. So this added logic only affects strict mode.

fisker (approved):
We are close to 100%, thanks for your hard work!

3cp:
The rest has some really hard ones like the "async of".

PR #331, opened 2024-08-29
feat(parser): support class auto-accessor through next option

### LINKED ISSUE #327: Missed part of Stage3 decorator spec
https://github.com/tc39/proposal-decorators?tab=readme-ov-file#introduction

The decorator spec introduced a new keyword "accessor" and a new class field auto-accessor.

```js
class Example {
  accessor myBool = false;
  @withDecorator accessor another = 1;
}
```

3cp:
closes #327


3cp:
Added a fix for BigInt.

fisker (approved):

Issue #337, opened 2024-09-13
async () problem


3cp:
```js
// This is valid, the async(foo = await 0) is a function call.
async function t() { async(foo = await 0) }

// This is invalid, await cannot be used in argument list of the async () => arrow function expression
async function t() { async(foo = await 0) => 1 }
```

Since meriyah doesn't do backtrack, it's troublesome when parsing "(foo = await 0)".
It looks like we needs another Flag.Await to hold it and report later. Something like `Flags.Octals` that we report when see "use strict": `function a() {"\7"; "use strict" }`.

3cp:
A related issue around duplicated proto.
```js
// FIXME:
// async() (not a async function) has different code path.
// Following should fail with duplicated proto.
async ({__proto__: a, __proto__: b});

// But following should pass, note this is object pattern,
// not object literal, duplicated proto works here.
async ({__proto__: a, __proto__: b}) => 1;
```

fisker:
The original issue has been fixed?

3cp:
It looks so. I vaguely recall some contributor fixed this.

fisker:
Ha,

```
async function t() { async(foo = await 0) => 1 }
```

is fixed, but

```
async (x = await (2)) => {};
```

is not...


Opened https://github.com/meriyah/meriyah/issues/620


Let's reopen this too, maybe need some tests.

3cp:
One issue left after #634 
`await` spelled with a unicode escape is still accepted

```js
async (a = aw\u{61}it => {}) => {}
```

3cp:
Fixed by #647

PR #338, opened 2024-09-14
fix(parser): fix import.meta check, fix await check

3cp:
There is still #337 not fixed. But I will live with it for now.


3cp:
Added another fix for duplicated proto in parenthesis.

3cp:
Rest in whitelist are bit hard, except json-invalid.js is out of scope for meriyah.
@fisker let me know if there is anything else you want before I release meriyah v6.

fisker:
> Rest in whitelist are bit hard

I think it's fine to ignore some edge cases.

> if there is anything else you want before I release meriyah v6.

I don't have things to do now.

fisker (approved):

fisker:
One second

fisker:
Maybe we can fix this https://github.com/meriyah/meriyah/blob/c9857c6f24f64a460cccbbc87249720ee0e8e30d/test/test262-parser-tests/unsupported-features.txt#L4 before release v6?

3cp:
That one is impossible with meriyah.

meriyah doesn't parse regexp syntax tree, we uses JS runtime to validate regexp. That feature is not implemented in V8 and browsers. So it always pops out as error (before it's widely implemented).

3cp:
I will hold v6 release for few weeks.

fisker:
I'm aware that we doesn't parse regular expressions, I thought Nide.js already support, sorry.

Anyway, if we are not going to Implement parse logic for it, how do you think the following options

1. Find a small lib for it.
1. Skip parsing (maybe under a `flag`).
1. Let user pass a function to parse.

So we will not relay on runtime

3cp:
Wait a minute, Chrome added support of it in May 2024.
https://caniuse.com/mdn-javascript_regular_expressions_named_capturing_group_duplicate_named_capturing_groups

Nodejs will have it soon with next v8.

fisker:
I still think we should not relay on runtime.

3cp:
I agree, but not going to do anything about it now. That's quite some work.

In addition parsing regexp will make meriyah considerably bigger I think.
The regexp-duplicate-named-groups (duplicated name is allowed in separate branches) requires additional lexical analysis.

fisker:
What's your opinion for these options https://github.com/meriyah/meriyah/pull/338#issuecomment-2351047474

3cp:
Probably 1 or write ourself.

3cp:
Looks quite a lot of code.
https://github.com/acornjs/acorn/blob/master/acorn/src/regexp.js

PR #341, opened 2024-09-15
feat: remove option "directives"

3cp:
BREAKING CHANGE: remove option "directives" as directives is in ECMA spec, and ESTree spec. No point to make it an optional feature.


fisker (approved):

3cp:
Mind me releasing v6?

fisker:
Go

3cp:
FYI, I have greatly simplified our gh-pages, removed all toolchain, use static html file with native esm loading.

https://github.com/meriyah/meriyah/tree/gh-pages

fisker:
Looks great!

3cp:
And use `@babel/code-frame` to highlight error location in code :-)

PR #345, opened 2024-10-18
chore: update and clean dependencies

fisker:


3cp:
Oh, more new test262 cases to be investigated:-)

3cp (approved):

PR #349, opened 2024-10-28
Fix another issue with octal template strings

ypapouski:
The parser fails to parse an octal string template when the tagged function contains arguments, for example:
```js
tag('a')`'\\00a0'`
```


3cp (approved):
Thx!

3cp:
v6.0.3 is released.

PR #361, opened 2025-01-02
fix(parser): fix lexical analysis on export of async func

### LINKED ISSUE #360: Asynchronous function exports are not working.
When in `module` goal and exporting an async function like so:
```js
export async function hey () {
    // This should work?
}
```
It says: `[3:1]: Cannot export a duplicate name 'hey'`

3cp:
closes #360


src/parser.ts
```diff
@@ -3580,10 +3579,6 @@ function parseExportDeclaration(
           tokenLine,
           tokenColumn
         );
-        if (scope) {
-          key = declaration.id ? declaration.id.name : '';
-          declareUnboundVariable(parser, key);
-        }
```
3cp, line 3586: parseFunctionDeclaration already does declareUnboundVariable. This is same for non-async function export in previous lines.

fisker (approved):

3cp:
Thx!

Issue #375, opened 2025-04-03
Distributed typings for `version` have an invalid import


giles-v:
I think the image conveys this best -- the source code is one level above the `package.json`, but when distributed, it's two.

<img width="720" alt="Image" src="https://github.com/user-attachments/assets/dba94326-28a4-414c-8f90-471133b19594" />

giles-v:
To be clear: the issue here is that the `../` directory traversal reaches into the `dist/` folder, which doesn't contain the `package.json` file (whereas in source code, it traverses into the project root, which has the `package.json` in).

3cp:
Yes, that's what I thought too. I changed the code to let d.ts to skip the reference of package.json.

giles-v:
@3cp to confirm this is definitely corrected by https://github.com/meriyah/meriyah/pull/377, closing this issue 👍

3cp:
Yes

PR #377, opened 2025-04-03
fix(parser): get/set/from/of are allowed as identifiers

### LINKED ISSUE #376: Parse failure for `await set()` - cannot use "await" as identifier inside an async function
```js
import { parseModule } from "meriyah";

let code = `
let foo = async () => {
    bar();
    await set();
};
`;

const result = parseModule(code);
console.log(result);
```

If I remove the call to `bar()`, or rename `set()` to something else (e.g. `setFn()`), it parses without error.

`meriyah@6.0.5`.

3cp:
closes #376

### DIFF
--- a/src/meriyah.ts
+++ b/src/meriyah.ts
@@ -4,6 +4,7 @@ import * as ESTree from './estree';
 // Current version
 import { version as pkgVersion } from '../package.json';
 
+// This bypass troublesome package.json in generated d.ts file.
 const version: string = pkgVersion;
 
 /**
--- a/src/meriyah.ts
+++ b/src/meriyah.ts
@@ -2,7 +2,9 @@ import { Context } from './common';
 import { parseSource, Options } from './parser';
 import * as ESTree from './estree';
 // Current version
-import { version } from '../package.json';
+import { version as pkgVersion } from '../package.json';
+
+const version: string = pkgVersion;
 
 /**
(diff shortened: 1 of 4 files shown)


src/meriyah.ts
```diff
@@ -2,7 +2,10 @@ import { Context } from './common';
 import { parseSource, Options } from './parser';
 import * as ESTree from './estree';
 // Current version
-import { version } from '../package.json';
+import { version as pkgVersion } from '../package.json';
+
+// This bypass troublesome package.json in generated d.ts file.
+const version: string = pkgVersion;
```
giles-v, line 8: I assume this is a fix for #375? Curious why -- the error I got from TypeScript was a failure to resolve the `../package.json` import, which will still happen here. Looks like my report wasn't specific enough.

3cp: This lets typescript to skip reference of package.json in generated d.ts file.
```js
import { Options } from './parser';
import * as ESTree from './estree';
declare const version: string;
export declare function parseScript(source: string, options?: Options): ESTree.Program;
export declare function parseModule(source: string, options?: Options): ESTree.Program;
export declare function parse(source: string, options?: Options): ESTree.Program;
export { Options, ESTree, version };
```

fisker (approved):

3cp:
v6.0.6 is released.

PR #382, opened 2025-04-10
ci: use yarn to install dependencies

fisker:
`npm i` is so [slow](https://github.com/meriyah/meriyah/actions/runs/14377174160/job/40312236071?pr=380#step:4:19), choose yarn since it can use directly on ci.


3cp:
If you want speed, bun is way faster :)

3cp (approved):

fisker:
> If you want speed, bun is way faster :)

I know, but need additional step.

> choose yarn since it can use directly on ci.

PR #384, opened 2025-04-11
feat: add `module-sync` condition

fisker:
[`module-sync`](https://nodejs.org/api/packages.html#conditional-exports)

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,9 @@
   "typings": "dist/src/meriyah.d.ts",
   "exports": {
     "types": "./dist/src/meriyah.d.ts",
-    "import": "./dist/meriyah.min.mjs",
-    "require": "./dist/meriyah.cjs"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -10,8 +10,9 @@
   "typings": "dist/src/meriyah.d.ts",
   "exports": {
     "types": "./dist/src/meriyah.d.ts",
-    "import": "./dist/meriyah.min.mjs",
-    "require": "./dist/meriyah.cjs"
+    "module-sync": "./dist/meriyah.mjs",
+    "require": "./dist/meriyah.cjs",
+    "default": "./dist/meriyah.mjs"
```
fisker, line 15: I changed it to the non-minified version, which is easier to debug, I think.

3cp (approved):

PR #386, opened 2025-04-11
chore: refactor generate-unicode script

fisker:


scripts/bundle.mjs
```diff
@@ -9,9 +9,9 @@ import * as ts from 'typescript';
 
 // eslint-disable-next-line n/no-unsupported-features/node-builtins
 const { dirname } = import.meta;
-const ENTRY = join('../src/meriyah.ts', dirname);
-const TSCONFIG = join('tsconfig.json', dirname);
-const DIST = join('../dis/', dirname);
+const ENTRY = join(dirname, '../src/meriyah.ts');
+const TSCONFIG = join(dirname, '../tsconfig.json');
+const DIST = join(dirname, '../dist/');
```
fisker, line 14: Mistake from #383

src/estree.ts
```diff
@@ -382,6 +373,7 @@ export interface PropertyDefinition extends _Node {
   type: 'PropertyDefinition';
   key: PrivateIdentifier | Expression;
   value: any;
+  // @ts-expect-error -- Unknown reason
   decorators?: Decorator[];
```
fisker, line 377: Don't know how to fix this

> error TS2687: All declarations of 'decorators' must have identical modifiers.

Seems caused by rollup types.

src/estree.ts
```diff
@@ -664,8 +656,15 @@ export interface MetaProperty extends _Node {
...
-export interface MethodDefinition extends MethodDefinitionBase {
+export interface MethodDefinition extends _Node {
   type: 'MethodDefinition';
+  key: Expression | PrivateIdentifier | null;
+  value: FunctionExpression;
+  computed: boolean;
+  static: boolean;
+  kind: 'method' | 'get' | 'set' | 'constructor';
+  // @ts-expect-error -- Unknown reason
+  decorators?: Decorator[];
```
fisker, line 667: Same as above

3cp (approved):

3cp:
Played latest code, no idea how the ts2687 pops out.

fisker:
It seems rollup defined a `PropertyDefinition`, remove it TS will not throw the error.

I can be wrong.

3cp:
That makes sense. I have a branch using bun test, have some issue with test262 (might be due to commonjs). I can try bun bundler too, it's much simpler, but it doesn't support UMD output yet.

fisker:
> I can try bun bundler too, it's much simpler, but it doesn't support UMD output yet.

I thought about bundling with esbuild too (bun uses a similar API), UMD support requires a plugin.
In Prettier, I wrote a plugin to make it work https://github.com/prettier/prettier/blob/main/scripts/build/esbuild-plugins/umd.js

PR #389, opened 2025-04-12
test: add production test

fisker:
Prevent breaking bundle script

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,14 +1,20 @@
 import { defineConfig } from 'vitest/config';
 
+const IS_CI = Boolean(process.env.CI);
+
 export default defineConfig({
   test: {
     globals: true,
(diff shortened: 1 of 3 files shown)


vitest.config.ts
```diff
@@ -1,14 +1,20 @@
...
+
 export default defineConfig({
   test: {
     globals: true,
     include: ['test/**/*.ts'],
-    exclude: ['test/test-utils.ts'],
+    exclude: [
+      'test/test-utils.ts',
+      // Skip production test on local
+      ...(IS_CI ? [] : ['test/production/production-tests.ts'])
```
3cp, line 12: Should this be other way around? Add production test when IS_CI.

3cp: Oh, my bad :)

3cp (approved):

PR #394, opened 2025-04-12
test: replace `assert` with `node:assert/strict`

fisker:
Actual changes in tests [`b1e1636` (#394)](https://github.com/meriyah/meriyah/pull/394/commits/b1e163614cf7b6fa151d39e7cce524daaaf5d8b9)


scripts/shared.mjs
```diff
@@ -0,0 +1,11 @@
...
+const supportsUnicodeSets = (() => {
+  try {
+    // eslint-disable-next-line n/no-unsupported-features/es-syntax
+    return new RegExp('', 'v').unicodeSets;
+  } catch {
+    return false;
+  }
+})();
+
+export { supportsUnicodeSets };
```
fisker, line 11: Extracting this because I thought `eslint-plugin-n` should have rule to forbid `assert`, turns out not implemented https://github.com/eslint-community/eslint-plugin-n/issues/59

I already extracted it, so I'm not going to change it back.

test/lexer/scanSingleToken.ts
```diff
@@ -84,19 +84,19 @@ describe('src/lexer/scan', () => {
...
+      t.equal(token, Token.NumericLiteral);
+      t.equal(state.tokenValue, Number(src));
     });
-    it('should recognise multi character numbers', () => {
+    it('should recognize multi character numbers', () => {
       const src = '345';
       const state = create(src, '', undefined);
       const token = scanSingleToken(state, Context.None, 0);
-      t.deepEqual(token, Token.NumericLiteral);
-      t.deepEqual(state.tokenValue, src);
```
fisker, line 99: We should update these `.deepEqual` to `.equal` in future, they are just primitive values.

test/lexer/template.ts
```diff
@@ -106,10 +106,10 @@ describe('Lexer - Template', () => {
   });
 
   describe('Lexer - Tagged Template', () => {
-    const tokens: [Context, Token, string, string | void][] = [
+    const tokens: [Context, Token, string, string | null][] = [
       //[Context.TaggedTemplate, Token.TemplateSpan, '`\\u{70bc`', undefined],
-      [Context.TaggedTemplate, Token.TemplateContinuation, '`\\7${', undefined],
-      [Context.TaggedTemplate, Token.TemplateContinuation, '`\\1${', undefined],
+      [Context.TaggedTemplate, Token.TemplateContinuation, '`\\7${', null],
```
fisker, line 111: Is the `tokenValue` expected to be null?

3cp: I guess so. Previously not strict assertion, so undefined == null I assume.

3cp (approved):

3cp (approved):

PR #397, opened 2025-04-14
chore: add spellcheck

fisker:


src/estree.ts
```diff
@@ -169,7 +169,10 @@ export type Expression =
   | UnaryExpression
   | UpdateExpression
   | YieldExpression;
-export type ForInitialiser = Expression | VariableDeclaration;
+export type ForInitializer = Expression | VariableDeclaration;
+// TODO: Misspelled, remove this in v7
+// spellchecker: disable-next-line
+export { ForInitializer as ForInitialiser };
```
fisker, line 175: Shuld we consider removing this as breaking change?

3cp: "Initializer" is US spelling, which is correct for Kenny.

3cp: cspell should have config for US and British.

fisker: This PR is changing "Initialiser" to "Initializer", I believe both US and British use "Initializer", I'm adding this alias, since it was exported.

3cp: Oh I missed that.

3cp (approved):

PR #399, opened 2025-04-14
chore: fix types for test files

fisker:
In VSCode, it can't find type declaration for vitest globals, eg `describe` and `it`.

`tsconfigOverride` in 'rollup-plugin-typescript2'  can't override `inlcude`, so I just added `tsconfig.json` for bundler.


3cp:
Better not to use vitest global. Use explicit vitest import like everybody normally does.

PR #431, opened 2025-06-11
test: use snapshot test

fisker:
~Waiting for #430 to merge first, to avoid conflicts.~

Fixes #429

This PR is huge. Suggest review commits one by one.

What I did is,

1. Added an ESLint rule to remove 3rd argument of `pass()`. [`6bd1d83` (#431)](https://github.com/meriyah/meriyah/pull/431/commits/6bd1d8319ba0dc7d2e5a2cf635eccc7875d70b28), this rule removed in [`ad3897b` (#431)](https://github.com/meriyah/meriyah/pull/431/commits/ad3897b37d726b0b5167900b80c72b8d024fc3bc)
2. Update `pass()` [`6dc711a` (#431)](https://github.com/meriyah/meriyah/pull/431/commits/6dc711af3ca0739e7546de1087337638e8046bad)
3. Other changes are by scripts.

PR #433, opened 2025-06-12
ci: setup pkg.new.pr

fisker:
https://pkg.pr.new/

To make it work, we need to install the GitHub application first.

https://github.com/apps/pkg-pr-new


fisker:
It works! https://github.com/meriyah/meriyah/pull/431#issuecomment-2969261575

PR #435, opened 2025-06-13
refactor: reuse `finishNode` in `parseSource`

fisker:
No logic change.


fisker:
Hold on this since it won't work with #436

PR #436, opened 2025-06-13
refactor: covert parser state to class

fisker:
Should we call it `State` or `Parser`?


3cp (approved):

3cp:
It looks ok. No idea why the parser and state are totally separate.

PR #441, opened 2025-06-13
refactor: remove unnecessary location parameters

### LINKED ISSUE #440: Remove `start,line,colum` from `parse*` function
They can be accessed from the parser state, I don't know why they are passed as arguments, which makes it hard to debug where they come from.

I'll start removing them from leaf nodes.

fisker:
In all these functions, I use

```
assert.deepEqual(
  {
    start,
    line,
    column,
  },
  {
    start: parser.tokenIndex,
    line: parser.tokenLine,
    column: parser.tokenColumn,
  },
);
```

to make sure the passed arguments are equal first, then remove them.

If every function is tested, there should be no logic changes.

Fixes #440


3cp (approved):

3cp:
This better works :)

PR #443, opened 2025-06-14
refactor: add `tokenStart` getter to `Parser`

fisker:
Use `tokenStart` instead of individual `tokenIndex`,`tokenLine`,`tokenColumn` which can easily get wrong, I see a few places already not correct.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -386,10 +386,10 @@ export function parseStatementList(
 
   while (parser.getToken() === Token.StringLiteral) {
     // "use strict" must be the exact literal without escape sequences or line continuation.
-    const { index, tokenValue, tokenStart } = parser;
+    const { index, tokenValue, tokenStart, tokenIndex } = parser;
     const token = parser.getToken();
     const expr = parseLiteral(parser, context);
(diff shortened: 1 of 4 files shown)


src/parser.ts
```diff
@@ -1327,23 +1240,14 @@ export function parseIfStatement(
...
   parser.assignable = AssignmentKind.Assignable;
-  const test = parseExpressions(
-    parser,
-    context,
-    privateScope,
-    0,
-    1,
-    parser.tokenIndex,
-    parser.line,
-    parser.tokenColumn,
```
fisker, line 1345: I guess this is a mistake.

src/parser.ts
```diff
@@ -7465,49 +6317,20 @@ export function parseObjectLiteralOrPattern(
...
 
-          value = parseMethodDefinition(
-            parser,
-            context,
-            privateScope,
-            state,
-            inGroup,
-            parser.tokenIndex,
-            tokenLine,
-            tokenColumn,
```
fisker, line 7509: I tink this is a mistake too.

3cp (approved):

Issue #421, opened 2025-06-06
Range of `ClassDeclaration` / `ClassExpression` should include decorators?


overlookmotel:
Similar to 2nd part of #416, and #417...

This is debatable, but in my opinion it'd be preferable if the range of a `ClassDeclaration` or `ClassExpression` included decorators. This would align with TS-ESLint and Babel.

Meriyah's current behavior means that the range of child (`Decorator`) is outside the range of its parent node (`ClassDeclaration` / `ClassExpression`), which seems unintuitive.

```js
@dec class C {}
const D = @dec class {};
```

[AST explorer](https://ast-explorer.dev/#eNpdjVsKwjAQRbcyzHdXUBEEdRfzM6SjVtJJyEMqIXs3tBZKP+85F05Biz2++cPRhNEn7NA3MEkYv/xqy7R1GcSAsRwjXKFUUuM0JrjBGXaq1BMp6UJk9i6kv7ivai8GeXC222HRreVaq5ACEE5uyFYIe0ghS7fCwPqUeIAqc9oQacX6Awx9Rnw=)

### Exports

What is even more debatable is what to do in these cases:

```js
@dec export class E {};
@dec export default class {};
```

This has been a matter of debate between TS-ESLint and Babel. I believe they currently have different answers:

```js
// TS-ESLint
@dec export class A {}
     ^^^^^^^^^^^^^^^^^ ExportNamedDeclaration
            ^^^^^^^^^^ Class
^^^^                   Decorator

// Babel
@dec export class A {}
^^^^^^^^^^^^^^^^^^^^^^ ExportNamedDeclaration
^^^^^^^^^^^^^^^^^^^^^^ Class
^^^^                   Decorator
```

However, from https://github.com/estree/estree/issues/315#issuecomment-2033157062 it sounds like TS-ESLint intends to switch to Babel's approach in future.

overlookmotel:
Note to maintainers: Sorry for the onslaught of issues!

Just to explain: I am one of the core team of [Oxc](https://github.com/oxc-project/oxc). We have recently seeking to align our AST with ESTree / TS-ESTree. We generate conformance tests by running other parsers on the complete set of Test262 and Typescript test fixtures. Initially we were aiming to align with Acorn and TS-ESLint parser, but we added Meriyah last week to generate tests for decorators.

At that point I compared the ASTs produced by Acorn and Meriyah for all Test262 cases. That is when I noticed the discrepancies that I've opened issues for today.

[...]

PR #452, opened 2025-06-15
fix(parser): include decorators in range of `Class{Expression,Declaration}`

fisker:
Part of #421

decorators before export not fixed.


test/parser/next/decorators.ts
```diff
@@ -363,5 +364,15 @@ describe('Next - Decorators', () => {
     ],
     [`class A { @dec\nx }`, Context.OptionsNext | Context.OptionsRanges | Context.OptionsLoc],
     [`class A { @dec\nx(){} }`, Context.OptionsNext | Context.OptionsRanges | Context.OptionsLoc],
+    ['@dec export class E {};', Context.OptionsNext | Context.Module | Context.OptionsRanges | Context.OptionsLoc],
+    [
+      '@dec export default class {};',
+      Context.OptionsNext | Context.Module | Context.OptionsRanges | Context.OptionsLoc,
+    ],
+    // ['export @dec class E {};', Context.OptionsNext | Context.Module | Context.OptionsRanges | Context.OptionsLoc],
```
fisker, line 372: @3cp bug? We can't parse this.

3cp (approved):

test/parser/next/decorators.ts
3cp: It's in spec.
> Class decorators may exclusively come before, or after, export/export default.

Let me know if you want to fix it here, or with a new PR.

fisker: https://github.com/meriyah/meriyah/issues/454

PR #457, opened 2025-06-16
fix(parser): include decorators in `ClassDeclaration`, `ExportDeclaration`, and `ExportDefaultDeclaration`

### LINKED ISSUE #421: Range of `ClassDeclaration` / `ClassExpression` should include decorators?

fisker:
Fixes #421


fisker:
@3cp , all issues that @overlookmotel reported are close now, we properly can release a new version.

@overlookmotel, feel free to open issues if you find anything wrong.

3cp:
I will wait for fisker's other pr before new release. Should not be long.
Great thanks for fisker's impressive works!

PR #458, opened 2025-06-16
refactor: use `location` in errors

fisker:


src/parser.ts
```diff
@@ -3421,22 +3417,22 @@ export function parseAwaitExpressionOrIdentifier(
 
   if (isIdentifier) {
     if (context & Context.InAwaitContext)
-      reportMessageAt(start, parser.startIndex, parser.startLine, parser.startColumn, Errors.InvalidAwaitAsIdentifier);
+      reportMessageAt(
+        start,
+        { index: parser.startIndex, line: parser.startLine, column: parser.startColumn },
```
fisker, line 3422: I'm still trying to understand the difference between `startIndex` and `index`. But before that, I'm not going to change the logic.

3cp:
```ts
export function nextToken(parser: ParserState, context: Context): void {
  parser.flags = (parser.flags | Flags.NewLine) ^ Flags.NewLine;
  // capture end of last token
  parser.startIndex = parser.index;
  parser.startColumn = parser.column;
  parser.startLine = parser.line;
  
  // then scan next token, which set parser.tokenIndex and set new parser.index
  parser.setToken(scanSingleToken(parser, context, LexerState.None));
}
```

3cp: We probably can refactor and pack index/line/column into a position (object or array of 3 integers).

3cp: Saw you had `Location` object.

3cp: Suggest to use Readonly on Location, so the code has to replace the object all together, not mutating its index/line/column. That makes `copy = parser.someLocation` safe.

fisker: So `startIndex` is actually "token end" position, I expected the same. So in other cases I think maybe it's wrong to use `index` in some places. We don't have a visualized error range to see the difference, I'll try to take a closer look when I have time.

fisker: > Suggest to use Readonly on Location

👍 Feel free to improve, I'm not near my laptop.

3cp: https://meriyah.github.io/meriyah/ our official playground does use babel code frame to draw the error range. You can get the source code (a single html file with no toolchain at all) and use your meriyah dist file.

Does pr.new expose individual files like jsdelivr? If so I can enhance our tool to allow custom version of meriyah build.

3cp: You may need to update the html code since the error shape is changed.

fisker: > Does pr.new expose individual files like jsdelivr?

No, but esm.sh support access the files. https://esm.sh/#docs

3cp:
I will wait for you finish this one before the release.

fisker:
I didn't change the logic, I think it can merge. I just trying to understand the parser better.

src/parser.ts
fisker: > since the error shape is changed

No, I didn't, just reduced the arguments passed. The public error still the same.

3cp (approved):

PR #462, opened 2025-06-18
test: refactor `pass`

fisker:
Change the test case to an object, use `options` instead of `context`.
Support `{only: true}` for easier to debug ONE case.


3cp (approved):
Love it!

PR #465, opened 2025-06-19
fix: forbid pass `module` in `parse{Script,Module}`

fisker:
I don't think this should be allowed.

```
> require('meriyah').parseScript('import.meta.url',{module:true})
{
  type: 'Program',
  sourceType: 'module',
  body: [ { type: 'ExpressionStatement', expression: [Object] } ]
}
>
```

### DIFF
--- a/src/meriyah.ts
+++ b/src/meriyah.ts
@@ -1,4 +1,3 @@
-import { Context } from './common';
 import { parseSource, Options } from './parser';
 import { type Program } from './estree';
 // Current version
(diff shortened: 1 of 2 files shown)


fisker:
Close this since we decide to add `sourceType`, https://github.com/meriyah/meriyah/issues/510

fisker:
Though we replaced `module` with `sourceType`, we should still forbid `sourceType` in `parse{Script,Module}()`.

3cp:
I thought you suggested we remove parseScript and parseModule, only keep parse?

3cp:
Ok

3cp (approved):

PR #469, opened 2025-06-20
fix(parser): fix column of escaped identifier

### LINKED ISSUE #468: Incorrect loc for escaped identifier
```
> require('meriyah').parse('a\\u0065',{ranges:true, loc:true}).body[0].expression
{
  type: 'Identifier',
  name: 'ae',
  start: 0,
  end: 7,
  range: [ 0, 7 ],
  loc: { start: { line: 1, column: 0 }, end: { line: 1, column: 1 } }
}
```

The range is correct, but the `loc.end.column` is wrong.

fisker:
Fixes #468

### DIFF
--- a/test/parser/declarations/__snapshots__/functions.ts.snap
+++ b/test/parser/declarations/__snapshots__/functions.ts.snap
@@ -3015,9 +3015,9 @@ exports[`Declarations - Function > Declarations - Functions (fail) > function fo
 `;
 
 exports[`Declarations - Function > Declarations - Functions (fail) > function foo(p\\u0061ckage) { "use strict"; } 1`] = `
-"SyntaxError [1:37-1:38]: Unexpected strict mode reserved word
+"SyntaxError [1:43-1:44]: Unexpected strict mode reserved word
 > 1 | function foo(p\\u0061ckage) { "use strict"; }
-    |                                      ^ Unexpected strict mode reserved word"
+    |                                            ^ Unexpected strict mode reserved word"
 `;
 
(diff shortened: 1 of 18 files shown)


test/parser/declarations/__snapshots__/functions.ts.snap
```diff
@@ -3015,9 +3015,9 @@ exports[`Declarations - Function > Declarations - Functions (fail) > function fo
 `;
 
 exports[`Declarations - Function > Declarations - Functions (fail) > function foo(p\\u0061ckage) { "use strict"; } 1`] = `
-"SyntaxError [1:37-1:38]: Unexpected strict mode reserved word
+"SyntaxError [1:43-1:44]: Unexpected strict mode reserved word
 > 1 | function foo(p\\u0061ckage) { "use strict"; }
-    |                                      ^ Unexpected strict mode reserved word"
+    |                                            ^ Unexpected strict mode reserved word"
```
fisker, line 3020: This seems incorrect(both original and new).

3cp: This is tricky. The code set a flag (for this one, it's `Flags.HasStrictReserved` I think) when parsing `p\\u0061ckage`, the string literal could be valid for non-strict mode. Then it passes function body for "directives", and it triggers error by checking parser.flags.

Note we only set flag when parsing the function arguments, not saving any position information, it's impossible to get back to the real range for the failure.

3cp (approved):

PR #467, opened 2025-06-20
test: add test to compare AST with Acorn

fisker:

PR #478, opened 2025-06-22
fix(parser): fix line/column where string literal contains `U+2028` or `U+2029` followed by `\`

fisker:
Turns out [#477](https://github.com/meriyah/meriyah/pull/477) is not corret, it should not count a new line when escaped...

### DIFF
--- a/src/lexer/string.ts
+++ b/src/lexer/string.ts
@@ -45,14 +45,12 @@ export function scanString(parser: Parser, context: Context, quote: number): Tok
         ret += String.fromCodePoint(char);
       }
       marker = parser.index + 1;
-    }
-
-    if (parser.index >= parser.end) report(parser, Errors.UnterminatedString);
-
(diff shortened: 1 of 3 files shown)


fisker:
Acutally, babel and acorn parse this differently...

Babel

```
          "loc": {
            "start": {
              "line": 1,
              "column": 1,
              "index": 1
            },
            "end": {
              "line": 1,
              "column": 5,
              "index": 5
            }
          },
```

Acorn

```
        "loc": {
          "start": {
            "line": 1,
            "column": 1
          },
          "end": {
            "line": 2,
            "column": 1
          }
        },
```

But we are wrong to say it ends on line 3, let's align with Acorn?

fisker:
@3cp This should be the last issue about ranges/loc, after this get merged #467 should be able to pass, we can release a new version.

3cp (approved):

PR #480, opened 2025-06-22
fix(parser): fix edge case on property definition parsing

### LINKED ISSUE #475: PropertyDefinition wrongly parsed
```js
class A {
  ["1"] = ()=>{}
  ["2"] = 3;
};
```

It should be parsed as two `PropertyDefinition`.

```js
class A {
  ["1"] = ()=>{}["2"] = 3;
          ^^^^^^^^^^^      MemberExpression
          ^^^^^^^^^^^^^^^^ AssignmentExpression
};
```

3cp:
closes #475

I vaguely recall I read in ecma spec that line break has some effect in class body, but I cannot find that piece of text in ecma now. I will try to find it before merging this one.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -87,6 +87,6 @@ export default [
     },
   },
   {
-    ignores: ['dist', 'src/unicode.ts', 'test262/test262'],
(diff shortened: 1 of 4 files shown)


3cp:
This fix is not ideal, see comment in the issue linked.

Issue #481, opened 2025-06-24
Test262 test cases which Meriyah cannot parse


overlookmotel:
Acorn can parse the following Test262 test cases, but Meriyah cannot:
 
* https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/import/import-attributes/json-named-bindings.js
* https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/statements/await-using/throws-if-initializer-not-object.js
* https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/statements/class/elements/syntax/valid/grammar-field-named-get-followed-by-generator-asi.js
* https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/statements/class/elements/syntax/valid/grammar-field-named-set-followed-by-generator-asi.js
* https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/statements/for-await-of/head-lhs-async.js
* https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/statements/using/throws-if-initializer-not-object.js

This is parsing with the following options:

```
meriyah.parse(code, {
    module: isModule,
    impliedStrict: isModule,
    ranges: true,
    raw: true,
    preserveParens: false,
    globalReturn: true,
    webcompat: true,
    next: true,
});
```

`isModule` depends on whether `module` appears in `flags` in the YAML header comment at top of each test case.

3cp:
The first one. https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/import/import-attributes/json-named-bindings.js

The description clearly says this is a syntax error, which meriyah correctly throws.
It's acorn that failed the spec.

The second one (explicit-resource-management) is not supported yet, as stated in our readme.

3cp:
Oh, the first one should parse.

```
negative:
  phase: resolution
  type: SyntaxError
```

I guess that `phase: resolution` is a step after parsing. Then we probably should remove that check in our code.

fisker:
* [x] https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/import/import-attributes/json-named-bindings.js https://github.com/meriyah/meriyah/pull/484
[...]

PR #484, opened 2025-06-27
fix(parser): stop complaining about named import/export with `{type: "json"}` attribute

fisker:
Ref https://github.com/meriyah/meriyah/issues/481#issuecomment-3000124591


3cp (approved):
Thx!

PR #487, opened 2025-06-28
chore: add `knip`

fisker:


src/chars.ts
```diff
@@ -12,19 +12,19 @@ export const enum Chars {
     FormFeed       = 0x0C,
     CarriageReturn = 0x0D,
 
-    Space       = 0x20,
+    // Space       = 0x20,
```
3cp, line 15: I am not convinced to comment out those unused chars. I imagine it would cause some trouble in auto-complete when we were implementing some few ecma feature that uses some of those chars.

fisker: Reverted [`dea83a5` (#487)](https://github.com/meriyah/meriyah/pull/487/commits/dea83a585cd6b46797e0003ddd69d063058f3897)

3cp (approved):

PR #490, opened 2025-06-29
refactor: extract `Scope`

fisker:
- Add `Scope` class
- Move scope related code around

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -28,18 +28,10 @@ import {
   validateAndDeclareLabel,
   HoistedClassFlags,
   HoistedFunctionFlags,
-  createScope,
-  addChildScope,
-  ScopeKind,
-  type ScopeState,
-  addVarName,
-  addBlockName,
   addBindingToExports,
   declareUnboundVariable,
   isEqualTagName,
   isValidStrictMode,
-  createArrowHeadParsingScope,
-  addVarOrBlock,
   isValidIdentifier,
   classifyIdentifier,
   type PrivateScopeState,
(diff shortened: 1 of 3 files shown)


src/parser.ts
```diff
@@ -6556,7 +6549,7 @@ function parseParenthesizedExpression(
 
   nextToken(parser, context | Context.AllowRegExp | Context.AllowEscapedKeyword);
 
-  const scope = context & Context.OptionsLexical ? addChildScope(createScope(), ScopeKind.ArrowParams) : void 0;
+  const scope = context & Context.OptionsLexical ? new Scope().createChildScope(ScopeKind.ArrowParams) : void 0;
```
fisker, line 6552: Not sure why this adds an empty parent scope, change to `new Scope(ScopeKind.ArrowParams)` the tests also passed, should I do it?

src/parser.ts
```diff
@@ -7410,7 +7403,7 @@ function parseAsyncArrowOrCallExpression(
 ): ESTree.CallExpression | ESTree.ArrowFunctionExpression {
   nextToken(parser, context | Context.AllowRegExp);
 
-  const scope = context & Context.OptionsLexical ? addChildScope(createScope(), ScopeKind.ArrowParams) : void 0;
+  const scope = context & Context.OptionsLexical ? new Scope().createChildScope(ScopeKind.ArrowParams) : void 0;
```
fisker, line 7406: Same here.

src/parser/scope.ts
```diff
@@ -0,0 +1,242 @@
...
+
+/**
+ * Create a parsing scope for arrow head, and add lexical binding
+ *
+ * @param parser Parser state
+ * @param context Context masks
+ * @param value Binding name to be declared
+ */
+export function createArrowHeadParsingScope(parser: Parser, context: Context, value: string): Scope {
+  const scope = new Scope().createChildScope(ScopeKind.ArrowParams);
```
fisker, line 93: And here.

src/parser.ts
3cp: addVarName, isPrivateIdentifierDefined also uses scope.parent.

3cp:
This also means we (and test262) don't have enough test coverage on the lexical analysis feature.

PR #500, opened 2025-07-01
refactor: add `Parser#{createScope,createScopeIfLexical}()`

fisker:
So we don't need pass `parser` around.


3cp (approved):

3cp:
Scope does sound like a state of parser.

3cp (approved):

PR #505, opened 2025-07-01
fix: fix `decodeHTMLStrict()`

fisker:
There is a bug when decoding named html entries, it try to access the object prototype, eg:

```
'&toString;' // -> 'function toString() { [native code] }'
```

I changed it to use `Object.hasOwn()`.

I also updated `descKeywordTable` to use the same way instead `Object.crate(null)`, so we can use less code.

I'm fine with alternative solutions:

1. Use `Map` instead
2. Use `{__proto__:null}`

Do you have a preference?


3cp (approved):

3cp:
LGTM

PR #516, opened 2025-07-03
feat!: generate unique node for shorthand `{Import,Export}Specifier`

fisker:
Should this be a `feat` or `fix`?


src/parser.ts
```diff
@@ -2321,7 +2321,7 @@ function parseImportSpecifierOrNamedImports(
       // by the keyword 'as'.
       // See the ImportSpecifier production in ES6 section 15.2.2.
       validateBindingIdentifier(parser, context, BindingKind.Const, token, 0);
-      local = imported;
+      local = parser.cloneIdentifier(imported);
```
fisker, line 2324: Should this be under `options.uniqueKeyInPattern`?

3cp: I guess so. But you are going to remove the option, so it's fine.

3cp:
Fix sounds better.

fisker:
Maybe people use `local === exported` to detect shorthand?

Let's ship it in v7, without option.

PR #547, opened 2026-02-05
chore: npm pkg delete is not necessary before npm publish

3cp:
Our package.json has "files" field, which tells "npm publish" what files to be included in the final npm package. You can verify this with "npm run build" then "npm package" (build the pkg without publish).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,5 +21,4 @@ jobs:
           cache-dependency-path: ./package.json
       - run: yarn
       - run: npm run build
-      - run: npm pkg delete scripts devDependencies lint-staged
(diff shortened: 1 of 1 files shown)


fisker:
> Our package.json has "files" field, which tells "npm publish" what files to be included in the final npm package.

Maybe you misunderstood what it does?

It deletes these fields in package.json, not delete directories or files.

> is not necessary

Yes, but these fields are not useful, shouldn't it nice to have them removed?

Check the diff in package.json file https://app.renovatebot.com/package-diff?name=meriyah&from=7.0.0&to=7.1.0

3cp:
:D thx! I did not understand it.

PR #549, opened 2026-02-07
fix(parser): fix for-of initializer check

### LINKED ISSUE #548: ParseError: [1:6-2:3]: 'for-of' loop head declarations can not have an initializer
Meriyah fails to parse the following JS snippet:
```js
var of=()=>{} \n of.prop;
```
with the error specified in a ticket title.

3cp:
closes #548


3cp:
This change reduced the chance of emitting a parser error. Since it didn't fail any unit tests or test262, I am not overly concerned of the fact that I don't quite understand it.

fisker (approved):
Should we add some tests?

PR #551, opened 2026-02-09
fix(parser): fix the invoke check after arrow function when newline is involved

### LINKED ISSUE #550: ParseError: [2:29-2:30]: Unexpected token: ']'
Meriyah fails to parse the following JS snippet:
```js
const a = () => { return 0; } \n
[{ foo: 'foo', },].forEach(() => {})
```
with the error specified in the title.

3cp:
closes #550

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -132,6 +132,7 @@ export const enum Flags {
   DisallowCall = 1 << 10,
   HasOptionalChaining = 1 << 11,
   EightAndNine = 1 << 12,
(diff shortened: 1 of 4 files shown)


3cp:
Both `() => {}\n[...` and `() => {}\n(...` have the same problem. Meriyah had a solution for `(`, I copied the same solution for `[`.

However I could not find relevant section in ecma spec for this.

3cp:
@fisker please have a look. Thx!

fisker (approved):

PR #554, opened 2026-03-25
Ranges option format

### LINKED ISSUE #553: Option to emit start/end without range array
## Problem

When `ranges: true`, meriyah emits three properties per node: `start`, `end`, and `range: [start, end]`. The `range`
array duplicates `start`/`end` and costs ~72 bytes per node in V8 (array object header + backing store + 2 tagged
slots).
On a baseline node, that's more than doubles the memory consumption. For large scripts this adds up significantly.

V8 per-node memory cost for reference:

| Shape                                       | Bytes |
|---------------------------------------------|-------|
| `{ type, name }`                            | 48B   |
| `{ type, name, start, end }`                | 64B   |
| `{ type, name, start, end, range: [s, e] }` | 136B  |

## Proposal

Add an option to control what format range information is emitted in:

```js
parse(code, {
    ranges: true,           // current behavior: start + end + range (backwards compatible)
    // or
    ranges: 'properties',   // only start/end properties, no range array
    // or
    ranges: 'array',        // only range array, no start/end
});
```

DimaIT:
Fixes https://github.com/meriyah/meriyah/issues/553

Implemented option suggestion from @3cp 

> The "properties" and "array" are not very intuitive to me. If we keep this approach (TBD), should we consider "start-end" and "range-array" (or just "range")? Or more detailed `ranges?: boolean | { start: boolean, end: boolean, range: boolean }` (I personally prefer this verbose version).

Also replaced rollup-plugin-typescript2 with @rollup/plugin-typescript to fix the build


3cp (approved):
Thx very much!

3cp:
@fisker pls have a look too.

fisker (approved):

PR #558, opened 2026-06-23
Fix the error range for strict reserved parameters

### LINKED ISSUE #470: Incorrect error range for `use strict` directive in function
```js
function a(package){'use strict'}
```

SyntaxError: [1:32-1:33]: Unexpected strict mode reserved word

Should report on `package` instead.

morgan-coded:
When a `"use strict"` directive appeared inside a function body, Meriyah could report a strict reserved parameter error near the end of the body instead of at the parameter itself.
The parser now keeps the first future-reserved parameter range while strictness is still unknown and uses that range if the later directive makes it invalid. I checked the red-to-green repro, 981 focused parser tests, the full suite with 94,304 passing, Unicode tests, lint, build, and `git diff --check`.
Fixes #470

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -3424,7 +3424,16 @@ function parseFunctionBody(
         }
       }
       if (parser.flags & Flags.StrictEvalArguments) parser.report(Errors.StrictEvalArguments);
-      if (parser.flags & Flags.HasStrictReserved) parser.report(Errors.UnexpectedStrictReserved);
+      if (parser.flags & Flags.HasStrictReserved) {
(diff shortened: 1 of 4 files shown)


3cp:
Thx very much! I will review it in few days.

3cp (approved):
Thx!

3cp:
v7.1.1 is released.

fisker:
@3cp Missed this page https://github.com/meriyah/meriyah/releases?

3cp:
Created release note.

PR #559, opened 2026-06-23
fix: accept \8 and \9 escapes in sloppy mode

spokodev:
## Problem

In sloppy mode under default options (`webcompat: false`), meriyah throws on `\8` and `\9` escapes in string literals, while accepting every other Annex B escape (`\7`, `\07`, `0777`, `08`, `09`).

ES2026 Annex B.1.2 places `\1`-`\7` (LegacyOctalEscapeSequence) and `\8`/`\9` (NonOctalDecimalEscapeSequence) in one `EscapeSequence` production, governed by a single early-error rule that is a Syntax Error only in strict mode. V8 (Node) and acorn both accept all of `\1`-`\9` in sloppy mode and reject them in strict mode. meriyah was uniquely gating `\8`/`\9` behind the `webcompat` option, diverging from that production.

## Fix

Drop the `!parser.options.webcompat` term in `src/lexer/string.ts` so `\8`/`\9` follow the same sloppy path as `\1`-`\7`. Strict-mode rejection still happens via the existing `context & Context.Strict` check, and template literals still reject via `isTemplate`.

## Behavior after fix

- sloppy `"\8"` / `"\9"` accepted (was rejected)
- strict `"\8"` / `"\9"` still rejected
- module (implicit strict) still rejects
- template `` `\8` `` still rejects

## Tests

Added positive sloppy-accept cases for `\8`, `\9`, `\9999`, `\9b` and removed the eight now-obsolete sloppy negative cases in `test/lexer/strings.ts` and `test/lexer/template.ts`. The strict negative cases are untouched.

Verified red before / green after: with the source line reverted the new positive cases fail; with the fix the full suite passes (94302 tests).

### DIFF
--- a/src/lexer/string.ts
+++ b/src/lexer/string.ts
@@ -218,7 +218,7 @@ export function parseEscape(parser: Parser, context: Context, first: number, isT
     // `8`, `9` (invalid escapes)
     case Chars.Eight:
     case Chars.Nine:
-      if (isTemplate || !parser.options.webcompat || context & Context.Strict) return Escape.EightOrNine;
+      if (isTemplate || context & Context.Strict) return Escape.EightOrNine;
       parser.flags |= Flags.EightAndNine;
     // fallthrough
     default:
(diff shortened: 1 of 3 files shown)


3cp:
Thx very much! I will review it in few days.

3cp (approved):
Thx very much! I am bit surprised that test262 doesn't have separate test case for sloppy mode.

3cp:
v7.1.1 is released.

PR #562, opened 2026-07-08
fix: reject empty unicode code point escape \u{}

spokodev:
A delimited unicode escape must contain at least one hex digit, so `\u{}` is not a valid escape. The lexer read it as code point 0, which made `"\u{}"` and an untagged `` `\u{}` `` parse as valid and gave a tagged template a `cooked` value of the null character instead of `null`. V8 and acorn both reject `\u{}`.

This requires at least one hex digit in the `\u{...}` form. The repo already had a disabled test for this exact case in `test/lexer/strings.ts`, which this re-enables.


3cp (approved):
Thx!

PR #563, opened 2026-07-08
fix: mark template literal with substitutions as invalid assignment target

spokodev:
A template literal is never a valid assignment or update target, but `parseTemplate` left `parser.assignable` set to whatever its last inner expression produced, so programs like `` `${a}` = 1 ``, `` `${a}` += 1 ``, `` `${a}`++ `` and `` for (`${a}` of b); `` were accepted. V8 and acorn reject these.

This sets the target kind to `Invalid` after the template is parsed, matching the no-substitution template path. Adds fail cases for the assignment, update and for-of forms.


3cp (approved):
Thx!

PR #564, opened 2026-07-08
fix: treat let as identifier before in and instanceof

spokodev:
At statement start in sloppy mode, `let in x` and `let instanceof x` are expressions where `let` is an identifier reference, which V8 and acorn accept. The disambiguation test matched on the keyword bit that `in` and `instanceof` share with binding identifiers, so the parser entered the variable declaration branch and reported the operator as an invalid binding name.

This skips the declaration branch when the token after `let` is a binary operator, mirroring the for-statement head, which already treats `let in` as an expression. Adds pass cases for both operators.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1534,7 +1534,7 @@ function parseLetIdentOrVarDeclarationStatement(
   const token = parser.getToken();
   let expr: ESTree.Identifier | ESTree.Expression = parseIdentifier(parser, context);
 
-  if (parser.getToken() & (Token.IsIdentifier | Token.IsPatternStart)) {
+  if (parser.getToken() & (Token.IsIdentifier | Token.IsPatternStart) && (parser.getToken() & Token.IsBinaryOp) === 0) {
     /* VariableDeclarations ::
      *  ('let') (Identifier ('=' AssignmentExpression)?)+[',']
      */
--- a/test/parser/declarations/let.ts
+++ b/test/parser/declarations/let.ts
@@ -579,6 +579,8 @@ describe('Declarations - Let', () => {
   ]);
 
   pass('Declarations - Let (pass)', [
+    'let in x',
+    'let instanceof x',
     'for (let {[x]: y = z} of obj);',
     '[x = true] = y',
     'let [,] = x;',
--- a/test/parser/declarations/__snapshots__/let.ts.snap
+++ b/test/parser/declarations/__snapshots__/let.ts.snap
@@ -4021,6 +4021,54 @@ exports[`Declarations - Let > Declarations - Let (pass) > let foo; 1`] = `
 }
 `;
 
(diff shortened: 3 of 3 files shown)


src/parser.ts
```diff
@@ -1534,7 +1534,7 @@ function parseLetIdentOrVarDeclarationStatement(
   const token = parser.getToken();
   let expr: ESTree.Identifier | ESTree.Expression = parseIdentifier(parser, context);
 
-  if (parser.getToken() & (Token.IsIdentifier | Token.IsPatternStart)) {
+  if (parser.getToken() & (Token.IsIdentifier | Token.IsPatternStart) && (parser.getToken() & Token.IsBinaryOp) === 0) {
```
3cp, line 1537: @fisker could you help to have a look here?

The issue is that keyword "in" and "instanceof" match Token.IsIdentifier.

```
Keyword                 = 1 << 12,
Contextual              = 1 << 13 | Keyword,
IsIdentifier            = 1 << 17 | Contextual,
```

Why we put all keywords as IsIdentifier?
**UPDATE**: never mind, my brain spins back now, keyword can be used as identifier `obj.in`.

~~I think `if ((parser.getToken() & Token.IsIdentifier) && (parser.getToken() & Token.IsPatternStart))` makes more sense here. (Tested, it works)~~ Brain fart :(

test/parser/declarations/let.ts
```diff
@@ -579,6 +579,8 @@ describe('Declarations - Let', () => {
   ]);
 
   pass('Declarations - Let (pass)', [
+    'let in x',
+    'let instanceof x',
```
3cp, line 583: Need to add these two cases to failed in strict mode too.

3cp (changes requested):

3cp:
Thx very much! I am confused about some meriyah internals, will have some discussion with @fisker.

src/parser.ts
3cp: TODO: line 1882 might need similar fix.

fisker:
Can't give suggestions due to my poor knowledge in parsers 😄 .

src/parser.ts
3cp: `Token.Reserved` is more precise than `Token.IsBinaryOp`, and it's reasonable in the context.

```js
if (parser.getToken() & (Token.IsIdentifier | Token.IsPatternStart) &&
  (parser.getToken() & Token.Reserved) !== Token.Reserved) {
```

It reads "if following token looks like an identifier or pattern, but not a reserved keyword".

3cp (changes requested):

3cp:
Superseded by #577.

Thank @spokodev for the work!

3cp:
v7.1.2 is released.

PR #566, opened 2026-07-10
chore: update ESLint to v10

fisker:


.github/workflows/node.js.yml
```diff
@@ -39,10 +39,14 @@ jobs:
...
           key: ${{ hashFiles('test262/test262-commit-hash') }}
       - run: yarn
+        if: matrix.node != '20'
+      - run: yarn --ignore-engines
+        if: matrix.node == '20'
       - run: npm run build
       - run: npm run test
       - run: npm run test:unicode
       - run: npm run lint
+        if: matrix.node != '20'
```
fisker, line 49: I hope it's fine to use `--ignore-engine` and skip lint on Node.js v20.

3cp: I don't mind. v20 is EOL. Althrough Meriyah doesn't have to drop support of v20, otherwise it's a major version change.

3cp (approved):

PR #571, opened 2026-07-10
chore: update rollup and plugins

fisker:
I also tried to switch to rolldown, but it seems too early, at least it can't output ASCII-only code.


package.json
```diff
@@ -81,10 +81,9 @@
     "@babel/code-frame": "^8.0.0",
     "@eslint/js": "^10.0.1",
     "@rollup/plugin-json": "^6.1.0",
-    "@rollup/plugin-terser": "^0.4.4",
+    "@rollup/plugin-terser": "^1.0.0",
     "@rollup/plugin-typescript": "^12.3.0",
-    "@stylistic/eslint-plugin": "^5.1.0",
-    "@types/babel__code-frame": "^7.0.6",
```
fisker, line 87: Bad merge in #566 😄

tsconfig.bundle.json
```diff
@@ -3,7 +3,8 @@
   "compilerOptions": {
     "rootDir": "src/",
     "declarationMap": false,
-    "declarationDir": "dist/types"
+    "declarationDir": "dist/types",
+    "sourceMap": false
```
fisker, line 7: Fixes warning

```
[plugin typescript] @rollup/plugin-typescript: Rollup 'sourcemap' option must be set to generate source maps.
```

3cp (approved):

PR #572, opened 2026-07-10
style: add explicit extension to imports

fisker:
I checked the bundled files, `.d.ts` files now keep the `.ts` in imports, I tested in TS, and it still works as before. Should be fine?  https://github.com/microsoft/TypeScript/issues/61037

```patch
>git diff --no-index old-meriyah.d.ts meriyah.d.ts
diff --git a/old-meriyah.d.ts b/meriyah.d.ts
index adf75f9..3775110 100644
--- a/old-meriyah.d.ts
+++ b/meriyah.d.ts
@@ -1,9 +1,9 @@
-import { type Program } from './estree';
-import { type Options } from './options';
+import { type Program } from './estree.ts';
+import { type Options } from './options.ts';
 declare const version: string;
 export declare function parseScript(source: string, options?: Omit<Options, 'sourceType'>): Program;
 export declare function parseModule(source: string, options?: Omit<Options, 'sourceType'>): Program;
 export declare function parse(source: string, options?: Options): Program;
 export { type Options, version };
-export type * as ESTree from './estree';
-export { isParseError, type ParseError } from './errors';
+export type * as ESTree from './estree.ts';
+export { isParseError, type ParseError } from './errors.ts';
```


fisker:
Worth mentioning, doing this because I hope the `.ts` files can run directly in Node.js at some point.

3cp:
I am using bun as much as I can. BTW bun test is fast :)

3cp (approved):

PR #574, opened 2026-07-10
docs: update feature support status

fisker:


README.md
```diff
@@ -26,19 +26,23 @@
 - No backtracking
 - Low memory usage
 
+### Not yet supported features:
+
+- [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management)
```
fisker, line 31: This is a finished proposal now.

README.md
```diff
@@ -26,19 +26,23 @@
...
 - [Decorators](https://github.com/tc39/proposal-decorators)
-- [JSON Modules](https://github.com/tc39/proposal-json-modules)
 
 ### Not yet supported stage 3 features:
 
-- [Explicit resource management](https://github.com/tc39/proposal-explicit-resource-management)
-- [Source phase import](https://github.com/tc39/proposal-source-phase-imports)
+- [Source Phase Imports](https://github.com/tc39/proposal-source-phase-imports)
+- [Deferring Module Evaluation](https://github.com/tc39/proposal-defer-import-eval)
+- [...and more](https://github.com/tc39/proposals#stage-3)
```
fisker: Should we list more features here or remove this link? There are many stage 3 proposals, but not every proposal needs parser support.

3cp: Probably no need. That's exactly what I checked before, I only listed features that needs parser support.

fisker: Removed, but added `Deferring Module Evaluation` which is new.

3cp (approved):

PR #576, opened 2026-07-11
fix: parse get and set fields before generators

### LINKED ISSUE #485: Not parsing class property correctly
These are valid

```
class A {
  get
  *a() {}
}

```

```
class B {
  static get
  *a() {}
}
```

https://github.com/tc39/test262/blob/823f4cfd2004cf410bc0344304e22e12d243e955/test/language/statements/class/elements/syntax/valid/grammar-field-named-get-followed-by-generator-asi.js

https://github.com/meriyah/meriyah/issues/481

morgan-coded:
Class fields named `get` or `set` followed by a line terminator and a generator method were treated as generator accessors and rejected. This recognizes the ASI boundary for static and instance fields while leaving actual accessors unchanged. I checked the focused class suite with 1,305 tests passing, then ran build, lint, and the full suite with 94,319 passing, including Test262. Fixes #485


fisker (approved):
Approving because the result looks good, need wait for @3cp to verify.

3cp (approved):
Thx!

3cp:
v7.1.2 is released.

PR #577, opened 2026-07-11
fix(parser): treat let as identifier before in and instanceof

3cp:
Supersede #564

@spokodev I have tested my proposed change. It looks working well.

### DIFF
--- a/test/conformance/ast-alignment-test.ts
+++ b/test/conformance/ast-alignment-test.ts
@@ -13,10 +13,6 @@ const notAlignedTests = new Set([
   'language/expressions/template-literal/tv-line-continuation.js',
   'language/expressions/template-literal/tv-line-terminator-sequence.js',
   'built-ins/String/raw/special-characters.js',
-
-  // https://github.com/meriyah/meriyah/issues/475
-  'staging/sm/Function/function-name-computed-01.js',
-  'staging/sm/Function/function-name-computed-02.js',
 ]);
 
 it(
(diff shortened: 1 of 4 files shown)


test/conformance/ast-alignment-test.ts
```diff
@@ -13,10 +13,6 @@ const notAlignedTests = new Set([
   'language/expressions/template-literal/tv-line-continuation.js',
   'language/expressions/template-literal/tv-line-terminator-sequence.js',
   'built-ins/String/raw/special-characters.js',
-
-  // https://github.com/meriyah/meriyah/issues/475
-  'staging/sm/Function/function-name-computed-01.js',
-  'staging/sm/Function/function-name-computed-02.js',
```
3cp, line 19: These were notified by test log that can be removed. Not related to this PR.

Need to go back to double check #475 again.

3cp: #475 issue doesn't occur in current main.

fisker (approved):
👍

test/conformance/ast-alignment-test.ts
fisker: I mean to throw an error here  https://github.com/meriyah/meriyah/blob/cad06053fc9ee16a2217f84ca46d45d9b25c17e7/test/conformance/ast-alignment-test.ts#L61  , but I don't why I didn't.

3cp: Create another PR if you want.

PR #579, opened 2026-07-17
feat: support explicit resource management declarations

### LINKED ISSUE #517: Add support for `using` and `await using`

morgan-coded:
**Problem I saw:** Meriyah skipped the Stage-4 Explicit Resource Management grammar for ES2027, so `using` and `await using` declarations were not part of the parser.

**What I changed:** I followed the parser-only plan from #517: `using` stays contextual, both declaration forms work in their permitted statement and loop-head positions with the required early errors, and ESTree emits `VariableDeclaration.kind` as `"using"` or `"await using"`.

**How I checked it:** Enabling the test262 feature moved 143 files (286 parser outcomes) from the red census to green with no new whitelist entries; the full suite passes with 94,356 tests. The two parser items called out in #481—`using/throws-if-initializer-not-object.js` and `await-using/throws-if-initializer-not-object.js`—now parse.

This is default-on for v7; the `next`-gate question in #517 has not received a maintainer answer. Runtime disposal behavior, `async of`, the options surface, and other unsupported test262 features stay out of scope.

Fixes #517

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,6 @@
 - No backtracking
 - Low memory usage
 
-### Not yet supported features:
-
 ## ESNext Stage 3 features
 
 ### Supported stage 3 features:
--- a/README.md
+++ b/README.md
@@ -28,8 +28,6 @@
 
 ### Not yet supported features:
 
-- [Explicit Resource Management](https://github.com/tc39/proposal-explicit-resource-management)
-
 ## ESNext Stage 3 features
 
 ### Supported stage 3 features:
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1947,7 +1947,8 @@ function parseVariableDeclarationList(
     declarationKind,
   );
   const list: ESTree.VariableDeclarator[] = [firstDeclaration];
-  const resourceNames = declarationKind ? new Set([(firstDeclaration.id as ESTree.Identifier).name]) : undefined;
+  const resourceNames =
+    declarationKind && parser.options.lexical ? new Set([(firstDeclaration.id as ESTree.Identifier).name]) : undefined;
   while (consumeOpt(parser, context, Token.Comma)) {
     bindingCount++;
     const declaration = parseVariableDeclaration(parser, context, scope, privateScope, kind, origin, declarationKind);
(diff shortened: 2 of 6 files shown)


fisker:
> the next-gate question in https://github.com/meriyah/meriyah/issues/517 has not received a maintainer answer.

It's answered

> we don't need to use "next" flag for stage 4 features.

README.md
```diff
@@ -28,8 +28,6 @@
 
 ### Not yet supported features:
```
fisker:
```suggestion
```

README.md
```diff
@@ -28,8 +28,6 @@
...
```
3cp: Remove this line please. We will add the section back when needed.

3cp (changes requested):

3cp:
Thx very much! That was fast.

3cp (approved):

fisker:
I've tested in Prettier with this change, everything works as expected. https://github.com/prettier/prettier/pull/19657 ❤️

fisker:
@3cp Should we allow

```
{
using f = foo(), f = foo();  
}
```

 when `lexical: false` ?

src/parser.ts
```diff
@@ -1752,14 +1934,29 @@ function parseVariableDeclarationList(
...
+  );
+  const list: ESTree.VariableDeclarator[] = [firstDeclaration];
+  const resourceNames = declarationKind ? new Set([(firstDeclaration.id as ESTree.Identifier).name]) : undefined;
   while (consumeOpt(parser, context, Token.Comma)) {
     bindingCount++;
-    list.push(parseVariableDeclaration(parser, context, scope, privateScope, kind, origin));
+    const declaration = parseVariableDeclaration(parser, context, scope, privateScope, kind, origin, declarationKind);
+    if (resourceNames) {
+      const { name } = declaration.id as ESTree.Identifier;
+      if (resourceNames.has(name)) parser.report(Errors.DuplicateBinding, name);
```
3cp, line 1957: @fisker reminded us that we only do this in lexical analysis. When lexical is off, don't keep track of resouceNames at all, so that parser will be lighter.

3cp (changes requested):

morgan-coded:
Good catch — as pushed, `using` duplicates errored even with `lexical: false`, which is inconsistent with the house precedent (`{ let f = 1, f = 2; }` parses fine in that mode). Aligned now: duplicate checking for `using` / `await using` gates on `lexical` like the other lexical kinds — the Stage-4 early error stays under `lexical: true`, parse-only mode skips it. Pushed with tests covering both modes.

fisker (approved):

3cp (approved):

3cp:
v7.2.0 is released.

PR #580, opened 2026-07-17
fix: normalize template carriage returns per TV/TRV

### LINKED ISSUE #460: `TemplateElement` `cooked` and `raw` incorrect when contain `\r`
```js
`\r\r\n`
```

I think `cooked` and `raw` in `TemplateElement` should not contain `\r`. However, I'm not completely sure of what the correct output is for for some of Test262's more obscure cases is e.g. `\\\r\n`.

Acorn appears to get this wrong too (or maybe I'm wrong!)

morgan-coded:
Normalizes `<CR>` and `<CR><LF>` to `<LF>` in TemplateElement `cooked` and `raw`, per ECMA-262 TV/TRV (TRV of LineTerminatorSequence is 0x000A). Fixes #460.

Root cause: the scanner's only CRLF branch (`src/lexer/template.ts:47-50`) could never fire — `advanceChar` pre-increments `parser.index`, so `parser.source.charCodeAt(parser.index)` compares the current CR against itself. Cooked therefore kept CR bytes verbatim (:56), and raw is a verbatim source slice (:65) with no TRV normalization.

Changes:
- CR handling in the scan loop: consume optional LF, emit `\n` into cooked, advance the line counter (lone CR previously didn't advance `parser.line` inside templates — the :52 predicate only matched LF).
- Raw normalization at both slice sites, matching acorn's `.replace(/\r\n?/g, "\n")`.
- Line-continuation: cooked still elides `\<CR><LF>`; raw keeps `\` + `\n`.
- Ranges/offsets are untouched — normalization is applied to values, never to source positions.

Tests: byte-exact fixtures in `test/parser/expressions/template.ts` for `\r` / `\r\n` in head/middle/tail spans, tagged and untagged, line continuations, and the three-backslash cases from the issue thread — each asserting cooked AND raw against acorn's output; plus the three test262 files cited in #460 as fixtures. No test262 whitelist change: the divergence is AST content, not pass/fail, so it surfaces through the #467 comparison harness and these unit fixtures.

Non-goals: LS/PS handling and the invalid-escape `cooked: null` semantics are unchanged; no grammar work.

### DIFF
--- a/src/lexer/template.ts
+++ b/src/lexer/template.ts
@@ -62,7 +62,9 @@ export function scanTemplate(parser: Parser, context: Context): Token {
   advanceChar(parser); // Consume the quote or opening brace
   parser.tokenValue = ret;
 
-  parser.tokenRaw = parser.source.slice(start + 1, parser.index - (token === Token.TemplateSpan ? 1 : 2));
+  parser.tokenRaw = parser.source
+    .slice(start + 1, parser.index - (token === Token.TemplateSpan ? 1 : 2))
+    .replace(/\r\n?/g, '\n');
 
   return token;
 }
--- a/src/lexer/template.ts
+++ b/src/lexer/template.ts
@@ -44,10 +44,10 @@ export function scanTemplate(parser: Parser, context: Context): Token {
         }
       }
     } else if (parser.index < parser.end) {
-      if (char === Chars.CarriageReturn && parser.source.charCodeAt(parser.index) === Chars.LineFeed) {
-        ret += String.fromCodePoint(char);
+      if (char === Chars.CarriageReturn && parser.source.charCodeAt(parser.index + 1) === Chars.LineFeed) {
         parser.currentChar = parser.source.charCodeAt(++parser.index);
       }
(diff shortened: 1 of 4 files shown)


3cp (approved):
Thx!

src/lexer/template.ts
```diff
@@ -62,7 +62,9 @@ export function scanTemplate(parser: Parser, context: Context): Token {
   advanceChar(parser); // Consume the quote or opening brace
   parser.tokenValue = ret;
 
-  parser.tokenRaw = parser.source.slice(start + 1, parser.index - (token === Token.TemplateSpan ? 1 : 2));
+  parser.tokenRaw = parser.source
+    .slice(start + 1, parser.index - (token === Token.TemplateSpan ? 1 : 2))
+    .replace(/\r\n?/g, '\n');
```
fisker, line 67: We already scanned the whole text, so we know if there are `\r`s? Maybe we can skip the replace call in some cases?

@3cp @morgan-coded

3cp: Sounds right :)

3cp:
v7.2.0 is released.

PR #583, opened 2026-07-17
perf(lexer): skip template CR normalization when no carriage returns present

morgan-coded:
Skips template raw-value CR normalization when the scan did not encounter a carriage return, following [fisker's suggestion](https://github.com/meriyah/meriyah/pull/580#discussion_r3599951105) and [3cp's agreement](https://github.com/meriyah/meriyah/pull/580#discussion_r3599984630). Cooked values, raw values, and the invalid tagged-template fallback stay unchanged.

The unchanged template suites pass with 2,786 tests, and the full suite passes with 94,371 tests. Non-goals: no other scanner changes, benchmarks, or test churn.

### DIFF
--- a/src/lexer/template.ts
+++ b/src/lexer/template.ts
@@ -13,6 +13,7 @@ export function scanTemplate(parser: Parser, context: Context): Token {
   const { index: start } = parser;
   let token: Token = Token.TemplateSpan;
   let ret: string | null = '';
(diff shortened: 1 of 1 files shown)


3cp:
No idea how to use the copilot to update the comment. It seems didn't touch this PR but claimed it did.

I will merged this first.

3cp (approved):

PR #585, opened 2026-07-21
fix: `import` removed from export map

timfish:
6.1.0 dropped the `import` condition that 5.0.0/6.0.0 had. While `module-sync`/`default` still point at the identical file, tools that don't yet recognize `module-sync` (confirmed: `@rollup/plugin-commonjs`, and by extension Nitro/Vinxi's standalone-output tracing) silently fail to resolve the package.

I am attempting to fix this in `@rollup/plugin-commonjs` but including `import` will fix this for users before the root cause fix trickles down.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   "exports": {
     "types": "./dist/types/meriyah.d.ts",
     "module-sync": "./dist/meriyah.mjs",
+    "import": "./dist/meriyah.mjs",
     "require": "./dist/meriyah.cjs",
     "default": "./dist/meriyah.mjs"
   },


fisker:
> tools that don't yet recognize module-sync (confirmed: @rollup/plugin-commonjs

Is there an issue link?

timfish:
So far I've opened a Nitro issue with a very minimal reproduction using `meriyah`:
- https://github.com/nitrojs/nitro/issues/4456

I haven't reported to Rollup because it might be related to how they use the plugin.

fisker:
Thanks, let's see how they repond first.

timfish:
I've narrowed this down to two seperate bugs, one with @vercel/nft and another with nitropack. 

Here is the other:
- https://github.com/nitrojs/nitro/issues/4468

This last one can be worked around via a `moduleSyncCatchall` setting which just tells nft to include all conditions.

3cp:
@fisker any downside if we add import? I don't mind it if it helps for bypassing bugs from other tools.

fisker:
> any downside if we add import?

There shouldn't be.

timfish:
This would give me great reason to push to upgrade to v7 and drop Node v18 support 🎉

3cp:
Node 18 was already dropped.

3cp (approved):

3cp:
v7.2.0 is released.

PR #587, opened 2026-07-23
fix(parser): reject assignments to optional chain members

### LINKED ISSUE #586: Forbid assign to optional chaining
Currently we forbid

```
a?.b = d;
```

but allows

```
a?.b.c = d;
```

emerardd:
Fixes #586.

Member parsing reset the assignment target to `Simple` after a normal `.` or
`[]` access, even when that member was still part of an optional chain. This
allowed assignments and update expressions such as `a?.b.c = d` and
`a?.b[c]++`.

The parser now preserves an invalid assignment target while optional chaining
is active. Regression coverage includes dot and computed members, plus
parenthesized optional chains that remain valid assignment targets.

Validation:

- optional-chaining suite: 1,036 tests passed
- TypeScript, ESLint, Prettier, and `git diff --check`
- full suite: 94,379 tests passed; two Test262 jobs could not clone their
  external corpus in the sandbox


src/parser.ts
```diff
@@ -4008,7 +4011,7 @@ function parseMemberOrUpdateExpression(
 
         consume(parser, context, Token.RightBracket);
 
-        parser.assignable = AssignmentTargetKind.Simple;
+        parser.assignable = restoreHasOptionalChaining ? AssignmentTargetKind.Invalid : AssignmentTargetKind.Simple;
```
3cp, line 4014: Did you miss some commit? This git diff didn't show where the `restoreHasOptionalChaining` was defined or set.

3cp (dismissed):
There seems to be some missing lines.

src/parser.ts
3cp: NVM, noticed restoreHasOptionalChaining is an existing feature :)

3cp (approved):

3cp:
v7.2.0 is released.

PR #588, opened 2026-07-25
feat(parser): support import defer and import source phases

### LINKED ISSUE #385: Support stage2.7 import defer
https://github.com/estree/estree/blob/master/stage3/defer-import-eval.md?plain=1

https://github.com/tc39/proposal-defer-import-eval

`import defer` only allows `import defer * as xx`, as the details of the namespace `xx` is deferred.

It looks it doesn't have too much implications on syntax level, should be easy to support.

morgan-coded:
Stage-3 `import defer` and `import source` were both on the test262 skip list, so neither phase form parsed. Per fisker's scope call on #385 this takes both, and covers `ImportExpression` as well as `ImportDeclaration`.

Changes:

- `import defer * as ns from "m"` → `ImportDeclaration` with `phase: "defer"` and a single `ImportNamespaceSpecifier`; `import source x from "m"` → `phase: "source"` with a single `ImportDefaultSpecifier`.
- `import.defer("m")` and `import.source("m")` → `ImportExpression` carrying the same `phase` values.
- `phase` follows the ESTree stage-3 fragments: explicit `null` on ordinary imports, with the key omitted entirely when `next` is off — the `decorators` idiom this parser already uses.
- `defer` and `source` stay contextual. `import defer from "m"`, `import source, * as ns from "m"`, `import {defer} from "m"` and friends still bind as before, and an escaped spelling is never a modifier. `import source source from "m"` and `import source from from "m"` follow the corpus (`source-phase-import/import-source-binding-name-2_FIXTURE.js`) rather than a simpler lookahead rule.
- Two new errors for a wrong specifier shape after either modifier.

How I checked it: dropping `import-defer`, `source-phase-imports` and `source-phase-imports-module-source` from `test262/unsupported-features.txt` unskips 502 tagged corpus files, and the parser corpus passes with no new whitelist entries. The full suite passes with 94,428 tests across 140 files. The Acorn alignment harness needed one addition — Acorn omits `phase` on ordinary imports, so `fixAcornAst` backfills `null` exactly as it already does for `decorators`.

Non-goals: no evaluation or runtime semantics, no `export defer` re-exports, and `import-assertions` stays on the skip list.

Fixes #385


fisker (approved):
Good job! Thank you!

fisker:
Can you update this section https://github.com/meriyah/meriyah#esnext-stage-3-features?

test/parser/next/import-phase.ts
```diff
@@ -0,0 +1,64 @@
...
+  { code: 'import defer {a} from "m";', options: nextModule },
+  { code: 'import source * as ns from "m";', options: nextModule },
+  { code: 'import source {a} from "m";', options: nextModule },
+  { code: 'import defer "m";', options: nextModule },
+  { code: 'import source "m";', options: nextModule },
+  { code: String.raw`import d\u0065fer * as ns from "m";`, options: nextModule },
+  { code: 'new import.defer("m")', options: nextModule },
+]);
+
+fail('Import Phase syntax with next disabled', [
```
3cp, line 73: Please add few cases for using "defer" and "source" as identifier when next is disabled. That's all!

3cp (changes requested):
Awesome work!

test/parser/next/import-phase.ts
```diff
@@ -0,0 +1,64 @@
...
+  { code: 'import source x from "m";', options: { ...nextModule, lexical: true } },
+]);
+
+pass('Next - Import Phase without lexical tracking', [
+  { code: 'import defer * as ns from "m";', options: { ...nextModule, lexical: false } },
+  { code: 'import source x from "m";', options: { ...nextModule, lexical: false } },
+]);
+
+pass('Next - Import Phase contextual identifiers', [
+  { code: 'import defer from "m";', options: nextModule },
```
3cp, line 30: I see you got "defer" and "source" as identifier in next mode. We can just duplicate those lines with next off.

morgan-coded:
Both addressed.

- README: Source Phase Imports and Deferring Module Evaluation moved to the supported stage-3 list; the not-yet-supported subsection is empty after the move, so it's removed.
- Tests: the contextual-identifier cases are duplicated with `next` off — `defer` and `source` stay ordinary identifiers, and the new snapshots pin `phase` as absent from every next-off AST. The two phase-syntax lines from that block (`import source source from "m";`, `import source from from "m";`) can't parse without `next`, so they're added to the invalid-forms-with-next-disabled block instead.

3cp (approved):

Issue #590, opened 2026-07-25
Refactor idea: expand `next` flag internally


fisker:
When moving stage 3 feature out of `next` option, it's hard to tell which part should be removed.

If we expand `next` to something like

```
options.next_importDefer
options.next_importSource
```

and use these flags in parsers and tests instead of `options.next` it should be easier to maintain.

fisker:
An alternative solution is use a feature list like `plugins` in babel https://github.com/babel/babel/pull/13043/changes#diff-353287936198b9465a88c7e8a7ba9499d980a071df7140c7e8a42f93f4c419e0R122

3cp:
That sounds doable.
Plugin might be overkill for meriyah, as we try to be light and fast.

fisker:
I'll work on it after I clean the next options from tests.

PR #591, opened 2026-07-25
test: remove useless `next` option in test (1/n)

fisker:
There are many `next` options in tests not doing anything.
I'm removing them, but I'd like to split them into multiple PRs.

Hope https://github.com/meriyah/meriyah/issues/590 can solve this problem.


fisker:
Should we disaable the AI review? https://github.com/meriyah/meriyah/settings/rules/19450527

3cp (approved):
Oh yes.

PR #598, opened 2026-07-26
refactor: expand `next: true` to different language features internally

### LINKED ISSUE #590: Refactor idea: expand `next` flag internally

fisker:
Fixes #590

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2536,9 +2536,9 @@ function parseImportDeclaration(
       const token = parser.getToken();
       const { tokenValue } = parser;
       const isPhaseDefer =
-        parser.options.features & Features.ImportDefer && (token & Token.IsEscaped) === 0 && tokenValue == 'defer';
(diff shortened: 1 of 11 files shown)


src/parser.ts
```diff
@@ -8243,7 +8248,7 @@ function parseClassExpression(
       id,
       superClass,
       body,
-      ...(parser.options.next ? { decorators } : null),
+      ...(parser.features ? { decorators } : null),
```
3cp: Is this a miss?

3cp (commented):
Just the one line in question.

src/parser.ts
fisker: Thanks! Did double check, still missed 😄  I'm old.

3cp (approved):

PR #599, opened 2026-07-26
chore: remove unused `tslib` dependency

fisker:


3cp:
Your PR is still in draft mode.

fisker:
It's required by the rollup plugin, I'll take another look later.

PR #604, opened 2026-07-27
perf(lexer): replace keyword table with Map and add length pre-filter

### LINKED ISSUE #601: Keyword lookup in scanIdentifier costs ~13% of parse time — Map + length pre-filter measures 4-9% faster
Profiling meriyah against real-world bundles put ~12.9% of parse time in the keyword lookup at the
end of `scanIdentifier` — `getOwnProperty(descKeywordTable, tokenValue)`, which is
`Object.hasOwn(o, k) ? o[k] : undefined` over a 59-key object literal, so two hash probes plus a
non-inlined builtin for every identifier token.

Replacing that table with a `Map` and adding a length pre-filter (keywords are 2-11 characters — the
same bound `scanIdentifierSlowCase` already applies, computed from the indices without touching the
string) measures faster on every corpus I tried:

    real-world bundles, no options   +4% to +9%
    real-world bundles, {loc,ranges} +3% to +8%
    test262 slice (5,304 files)      +4% to +6%

I'm giving ranges rather than one number deliberately: I measured three independent sweeps and got
+9.4%, +7.4% and +3.9% on the plain real-world suite. Every sweep favoured the change (sign counts
15/15, 13/15, 12/15 paired samples), so I'm confident in the direction, but my machine is a loaded
8 GB M1 and the spread is wider than I'd like. The harness is 15 samples x 5 iterations, paired
median with a sign test, A/A noise floor measured at ±1.3%. Corpora: typescript 5.9.3, three, d3,
lodash and vue production bundles (10.9 MB total) plus a 5,304-file test262 slice. Happy to hand you
the harness so you can measure it on yours — I'd rather you saw your own number than took mine.

Diff is 2 files, +73/-66. `TEST262=1 npx vitest run` passes 143 files / 91,298 tests with the
whitelist unchanged, and I ran a differential check over 22,640 input x option-set pairs (the
corpora, the full test262 slice, and ~120 probes at the changed locus) confirming identical ASTs and
identical errors. Bundle cost is +296 bytes.

Two things I measured and am NOT proposing, in case they save you time later:

- Rewriting `scanSingleToken`'s dispatch over a dense key: no effect, inside the noise floor. That
  22% is per-character scanning work, not dispatch cost.
- Memoising the throwaway `{index, line, column}` object allocated at ~314 call sites: measurably
  slower. Three integer comparisons cost more than V8's bump-pointer allocation plus scavenge, so the
  GC share is mostly the `loc` objects the caller asked for, not removable overhead.

Would you like this as a PR?

morgan-coded:
Implements #601: the 59-key `descKeywordTable` object literal becomes a `Map`, and a length pre-filter (keywords are 2-11 characters, bounds computed from the scan indices without touching the string) short-circuits the lookup at the end of `scanIdentifier`.

Two commits — the Map swap, then the pre-filter. 2 files, +73/-66, +296 bytes on the minified ESM bundle.

Benchmark ranges and method are in #601 (+4% to +9% real-world, +3% to +8% with `{loc, ranges}`, +4% to +6% on the test262 slice); the harness offer there stands. Decomposition of the two halves and the measured `{__proto__: null}` comparison are in the issue thread, along with an independent replication on a second machine (i9-14900K, Windows 11, node 24) at +1.9% to +7.7%.

`TEST262=1 npx vitest run` passes 143 files / 91,298 tests with the whitelist unchanged, and the 22,640-pair differential check confirms identical ASTs and identical errors — both re-run at this exact head. The full conformance suite and the differential also pass on the Windows machine.

Closes #601.


src/lexer/identifier.ts
```diff
@@ -19,9 +18,17 @@ export function scanIdentifier(parser: Parser, context: Context, isValidAsKeywor
...
-    : scanIdentifierSlowCase(parser, context, 0, isValidAsKeyword);
+  if (parser.currentChar === Chars.Backslash || parser.currentChar > 0x7e) {
+    return scanIdentifierSlowCase(parser, context, 0, isValidAsKeyword);
+  }
+
+  // Every entry in `descKeywordTable` is between 2 and 11 characters long, so
+  // anything outside that range cannot be a keyword and can skip the lookup.
+  // Same bound as the escaped path below.
+  const length = parser.index - parser.tokenIndex;
+  if (length < 2 || length > 11) return Token.Identifier;
```
3cp: Should we pre-calculate the length boundary from descKeywordTable? It's better for future proof than hard coded values.

src/lexer/identifier.ts
```diff
@@ -86,7 +93,7 @@ export function scanIdentifierSlowCase(
 
   const { length } = parser.tokenValue;
   if (isValidAsKeyword && length >= 2 && length <= 11) {
```
3cp: Same here for the old code.

3cp (commented):
Just some thought on the 2,11 boundary.

morgan-coded:
Both addressed, plus the Map conversion from #601.

- Bounds derive from the table now — `minKeywordLength` / `maxKeywordLength` computed from `descKeywordTable.keys()` at module load, used in the fast path and in the escaped path. They evaluate to the same 2 and 11, and an A/B against the hardcoded version lands at chance (8/15 and 8/15 paired samples, inside the ±1.3% noise floor), so the indirection is free.
- `entities` and `decodeMap` in `decodeHTML.ts` are Maps as well, which retires `getOwnProperty` and deletes `utilities.ts`. One wrinkle worth flagging: `decodeMap` is looked up by numeric code point and `Object.entries` stringifies keys, so it is built as a `Map<number, number>` with the keys converted back — a string-keyed Map would silently miss every C1 remap. Checked with a 23-probe entity differential (named, decimal, hex, the C1 range, malformed and unterminated forms, each as JSX text and as an attribute value, across three option sets): identical output.

143 files / 91,298 tests at this head, and the 22,640-pair AST differential is unchanged. Bundle is +397 B over main in total.

3cp (approved):

3cp:
@fisker any other consideration?

fisker (approved):

PR #612, opened 2026-07-27
fix(parser): restrict decorator expressions to the decorators grammar

### LINKED ISSUE #407: Computed class property with decorator missing `key`
```js
  class A{
    @decorators
    [x] = 1
  }
```

It seems valid in TS, but I'm not sure, when playing with [this example](https://github.com/tc39/proposal-decorators/issues/556), I found meriyah emits a `PropertyDefinition` with `key: null`, it can't be right.

### LINKED ISSUE #408: Decorators incorrectly parsed
```
  class A{
    @decorators[0]
    method() {}
  }
```

Should  parse  as computed `0` property with `decorators` as decorator.

See https://github.com/tc39/proposal-decorators/issues/556

morgan-coded:
Fixes #408, fixes #407.

`parseDecoratorList` now implements the decorator grammar directly instead of running the general member/call machinery: an identifier head, `.` identifier / `.#private` steps, and at most one terminal argument list, with `@( Expression )` unchanged as the escape hatch.

Unparenthesized `@dec[0]`, `@dec["x"]`, `@dec?.b`, `@dec?.()`, `@dec().b`, `@dec()()`, and `@this.x` are now syntax errors — babel rejects all seven. The tokens freed by stopping at `[` flow to the class-element parser, so the multi-line forms from #408 parse the way babel/flow/TS/hermes parse them (`@decorators` attaches to a computed field `[0]`; the method is a separate, undecorated element), and #407's repro yields a `PropertyDefinition` with `computed: true` and key `x` instead of `key: null`.

Tests: fail cases for all seven forms plus the one-line `class A { @dec[0] m() {} }` (no line terminator, so the computed field cannot end before `m`), full AST assertions for the split shapes and the #407 repro, and controls keeping dotted chains, single calls, private-name steps, and every parenthesized form green. `TEST262=1 npx vitest run`: 143 files / 91,310 tests, whitelist untouched.

Downstream note: prettier's format suite currently disables `js/decorators/member-expression.js` for meriyah over exactly this divergence — with this change that fixture parses structurally identical to babel, so the disable can be dropped after the next release.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -8288,9 +8288,78 @@ function parseDecoratorList(
   nextToken(parser, context | Context.AllowRegExp);
   const expressionStart = parser.tokenStart;
 
-  let expression = parsePrimaryExpression(parser, context, privateScope, BindingKind.Empty, 0, 1, 0, 1, start);
+  let expression: ESTree.Decorator['expression'];
 
-  expression = parseMemberOrUpdateExpression(parser, context, privateScope, expression, 0, 0, expressionStart);
+  if (parser.getToken() === Token.LeftParen) {
+    expression = parsePrimaryExpression(parser, context, privateScope, BindingKind.Empty, 0, 1, 0, 1, start);
+  } else {
+    const token = parser.getToken();
+
+    if (
+      ((token & Token.IsIdentifier) !== Token.IsIdentifier && !isValidIdentifier(context, token)) ||
+      (context & Context.Strict && (token & Token.FutureReserved) === Token.FutureReserved)
+    ) {
+      parser.report(Errors.UnexpectedToken, KeywordDescTable[token & Token.Type]);
+    }
+
+    if (token === Token.AwaitKeyword && context & (Context.InAwaitContext | Context.Module)) {
+      parser.report(Errors.AwaitIdentInModuleOrAsyncFunc);
+    }
+
+    if (token === Token.YieldKeyword && context & Context.InYieldContext) {
+      parser.report(Errors.DisallowedInContext, 'yield');
+    }
+
+    let memberExpression: ESTree.Identifier | ESTree.MemberExpression = parseIdentifier(
+      parser,
+      context | Context.TaggedTemplate,
+    );
+
+    parser.assignable =
+      context & Context.Strict && (token & Token.IsEvalOrArguments) === Token.IsEvalOrArguments
+        ? AssignmentTargetKind.Invalid
+        : AssignmentTargetKind.Simple;
+
+    while (parser.getToken() === Token.Period) {
+      nextToken(parser, (context | Context.AllowEscapedKeyword | Context.InGlobal) ^ Context.InGlobal);
+
+      const property = parsePropertyOrPrivatePropertyName(parser, context | Context.TaggedTemplate, privateScope);
+
+      parser.assignable = AssignmentTargetKind.Simple;
+
+      memberExpression = parser.finishNode<ESTree.MemberExpression>(
+        {
+          type: 'MemberExpression',
+          object: memberExpression,
+          computed: false,
+          property,
+          optional: false,
+        },
+        expressionStart,
+      );
+    }
+
+    expression = memberExpression;
+
+    if (parser.getToken() === Token.LeftParen) {
+      const args = parseArguments(parser, context, privateScope, 0);
+
+      parser.assignable =
+        !(context & Context.Strict) && parser.options.webcompat
+          ? AssignmentTargetKind.WebCompat
+          : AssignmentTargetKind.Invalid;
+
+      expression = parser.finishNode<ESTree.CallExpression>(
+        {
+          type: 'CallExpression',
+          callee: memberExpression,
+          arguments: args,
+          optional: false,
+        },
+        expressionStart,
+      );
+    }
+  }
 
   return parser.finishNode<ESTree.Decorator>(
     {
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -193,6 +193,9 @@ function parseModuleItem(parser: Parser, context: Context, scope: Scope | undefi
       moduleItem = parseExportDeclaration(parser, context, scope);
       break;
     case Token.ImportKeyword:
+      if (parser.leadingDecorators.decorators.length) {
+        parser.report(Errors.InvalidLeadingDecorator);
+      }
       moduleItem = parseImportDeclaration(parser, context, scope);
       break;
     default:
@@ -238,6 +241,10 @@ function parseStatementListItem(
   // LexicalDeclaration[In, Yield] :
   //   LetOrConst BindingList[?In, ?Yield] ;
   const start = parser.tokenStart;
+
+  if (parser.leadingDecorators.decorators.length && parser.getToken() !== Token.ClassKeyword) {
+    parser.report(Errors.InvalidLeadingDecorator);
+  }
 
   switch (parser.getToken()) {
     //   HoistableDeclaration[?Yield, ~Default]
@@ -2951,6 +2958,15 @@ function parseExportDeclaration(
   // https://tc39.github.io/ecma262/#sec-exports
   nextToken(parser, context | Context.AllowRegExp);
 
+  if (
+    parser.leadingDecorators.decorators.length &&
+    parser.getToken() !== Token.DefaultKeyword &&
+    parser.getToken() !== Token.Decorator &&
+    parser.getToken() !== Token.ClassKeyword
+  ) {
+    parser.report(Errors.InvalidLeadingDecorator);
+  }
+
   const specifiers: ESTree.ExportSpecifier[] = [];
 
   let declaration: ESTree.ExportDeclaration | ESTree.Expression | null = null;
@@ -2961,6 +2977,14 @@ function parseExportDeclaration(
     // export default HoistableDeclaration[Default]
     // export default ClassDeclaration[Default]
     // export default [lookahead not-in {function, class}] AssignmentExpression[In] ;
+
+    if (
+      parser.leadingDecorators.decorators.length &&
+      parser.getToken() !== Token.Decorator &&
+      parser.getToken() !== Token.ClassKeyword
+    ) {
+      parser.report(Errors.InvalidLeadingDecorator);
+    }
 
     switch (parser.getToken()) {
       // export default HoistableDeclaration[Default]
@@ -8104,6 +8128,9 @@ function parseClassDeclaration(
   } else {
     start = parser.tokenStart;
     decorators = parseDecorators(parser, context, privateScope);
+    if (decorators.length && parser.getToken() !== Token.ClassKeyword) {
+      parser.report(Errors.InvalidLeadingDecorator);
+    }
   }
 
   context = (context | Context.InConstructor | Context.Strict) ^ Context.InConstructor;
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -8265,7 +8265,7 @@ function parseDecorators(parser: Parser, context: Context, privateScope: Private
 
   if (parser.features & Features.Decorators) {
     while (parser.getToken() === Token.Decorator) {
-      list.push(parseDecoratorList(parser, context, privateScope));
+      list.push(parseDecorator(parser, context, privateScope));
     }
   }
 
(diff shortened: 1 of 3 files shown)


fisker (approved):
❤️

test/parser/next/__snapshots__/decorators.ts.snap
```diff
@@ -66,6 +66,42 @@ exports[`Decorators > Decorators (fail) > @bar; 1`] = `
     |     ^ Class declaration must have a name in this context"
 `;
 
+exports[`Decorators > Decorators (fail) > @dec()() class A {} 1`] = `
+"SyntaxError [1:7-1:8]: Class declaration must have a name in this context
+> 1 | @dec()() class A {}
+    |        ^ Class declaration must have a name in this context"
```
3cp: I will have a play with this implementation in coming days. The error here (and the rest) is not ideal.

3cp:
Thank you for the work! I will take some days to play with it.

test/parser/next/__snapshots__/decorators.ts.snap
3cp: When I ran your branch, I got a slightly better error message, don't know why your snapshot message is different. It's correct message in its context, but we didn't error out early for "Leading decorators must be attached to a class declaration".

```
SyntaxError [1:9-1:14]: Expected '=>'
> 1 | @dec()() class A {}
    |          ^^^^^ Expected '=>'
```

```
SyntaxError [1:13-1:14]: Leading decorators must be attached to a class declaration
> 1 | @dec()() => {}
    |              ^ Leading decorators must be attached to a class declaration
```

It should be possible to error out as early as possible on first token after `@dec()` which is the following `(`. If leadingDecorators is not empty, the following token in a statement (parseStatementListItem) or export declaration (parseExportDeclaration) must be `class`. Note our leadingDecorators were parsed before class declaration, that was designed to cater `@dec.before.export export class A {}`.

src/parser.ts
3cp: The existing func name and doc is miss-leading. This func only parses single decorator. Please help to clean up too.

src/parser.ts
```diff
@@ -8288,9 +8288,78 @@ function parseDecoratorList(
...
 
-  expression = parseMemberOrUpdateExpression(parser, context, privateScope, expression, 0, 0, expressionStart);
+  if (parser.getToken() === Token.LeftParen) {
+    expression = parsePrimaryExpression(parser, context, privateScope, BindingKind.Empty, 0, 1, 0, 1, start);
+  } else {
+    const token = parser.getToken();
+
+    if (
+      ((token & Token.IsIdentifier) !== Token.IsIdentifier && !isValidIdentifier(context, token)) ||
+      (context & Context.Strict && (token & Token.FutureReserved) === Token.FutureReserved)
```
3cp, line 8315: Strict check is not needed. Decorator is only for class and only in module mode. Strict mode is ensured, but our test case setup needs to be careful.

3cp: I think I am wrong. Script mode still can have class and decorator.

src/parser.ts
```diff
@@ -8288,9 +8288,78 @@ function parseDecoratorList(
...
+
+    let memberExpression: ESTree.Identifier | ESTree.MemberExpression = parseIdentifier(
+      parser,
+      context | Context.TaggedTemplate,
+    );
+
+    parser.assignable =
+      context & Context.Strict && (token & Token.IsEvalOrArguments) === Token.IsEvalOrArguments
+        ? AssignmentTargetKind.Invalid
+        : AssignmentTargetKind.Simple;
```
3cp: I don't understand why we need this. I changed this code to `parser.assignable = AssignmentTargetKind.Invalid;` and also two more `parser.assignable =` in following lines. No test failed after the change.

3cp: Remove the three lines of `parser.assignable =` also worked without breaking anything. I think we can remove them. I could not think of any test cases for them.

`@dec(a=1) class A {}` works fine with those lines.

3cp (changes requested):

src/parser.ts
3cp, line 8301:
```suggestion
  // Decorator[Yield, Await] :
  //    @ DecoratorMemberExpression[?Yield, ?Await]
  //    @ DecoratorParenthesizedExpression[?Yield, ?Await]
  //    @ DecoratorCallExpression[?Yield, ?Await]
  //  DecoratorMemberExpression[Yield, Await] :
  //    IdentifierReference[?Yield, ?Await]
  //    DecoratorMemberExpression[?Yield, ?Await] . IdentifierName
  //    DecoratorMemberExpression[?Yield, ?Await] . PrivateIdentifier
  //  DecoratorParenthesizedExpression[Yield, Await] :
  //    ( Expression[+In, ?Yield, ?Await] )
  //  DecoratorCallExpression[Yield, Await] :
  //    DecoratorMemberExpression[?Yield, ?Await] Arguments[?Yield, ?Await]
  const start = parser.tokenStart;
```

Let's add the spec there.

3cp (changes requested):

test/parser/next/__snapshots__/decorators.ts.snap
3cp: Now I understand why the snap error is different from my test. I tested it with normal `parse()` call with module mode, but the unit test setup is:
```
{ code: '@dec()() class A {}', options: { features: Features.Decorators } },
```
That tests the code in script mode, the decorator is parsed by parseClassDeclaration, but my test parsed them in the leadingDecorators. Adding `options: { sourceType: 'module'` to the test case should yield the error message I saw locally.

Note the error message in script mode parseClassDeclaration is clearly not correct.

We need to fix the error message in both code paths.

3cp:
Screenshot 2026-07-28 at 2 28 47 pm

I updated our playground https://meriyah.github.io/meriyah to allow testing against a pkg.pr.new from a PR.

morgan-coded:
Implemented the revision, split by concern.

- Both paths now report at the offending token — your script/module read was right. `@dec()()` stops at the second `(` with `Leading decorators must be attached to a class declaration` in script and module mode alike, and paired test cases pin both.
- Renamed `parseDecoratorList` to `parseDecorator` and fixed its doc comment; `parseDecorators` keeps the list role.
- Dropped the three `parser.assignable` writes. Traced the four `parseDecorators` callers first — none read the value before a later parse overwrites it.
- Added the grammar productions above `parseDecorator`.
- Left the strict check alone: your retraction was correct. Script mode reaches decorators, which is also why the snapshot differed from your run.

One consequence worth flagging: erroring early moves the range on ten pre-existing fail snapshots (`@bar const foo = 1;`, `@bar export * from "./foo";`, `@bar function foo() {}` and similar). Each now points at the first token that cannot attach the decorators, instead of a later and less specific error.

At `e142129`: 143 files / 91,313 tests, plus lint, build and diff-check green. The pkg.pr.new build refreshes on this push, so the playground picks up the fix.

src/parser.ts
```diff
@@ -2951,6 +2958,15 @@ function parseExportDeclaration(
   // https://tc39.github.io/ecma262/#sec-exports
   nextToken(parser, context | Context.AllowRegExp);
 
+  if (
+    parser.leadingDecorators.decorators.length &&
+    parser.getToken() !== Token.DefaultKeyword &&
+    parser.getToken() !== Token.Decorator &&
```
3cp: Don't need this line. `@d export @e class A {}` will be an error anyway, let parseClassDeclaration handle that.

src/parser.ts
```diff
@@ -2962,6 +2978,14 @@ function parseExportDeclaration(
     // export default ClassDeclaration[Default]
     // export default [lookahead not-in {function, class}] AssignmentExpression[In] ;
 
+    if (
+      parser.leadingDecorators.decorators.length &&
+      parser.getToken() !== Token.Decorator &&
```
3cp: Same as above.

src/parser.ts
3cp: parseClassDeclaration already handles that.

src/parser.ts
3cp: (I had trouble to use github to suggest change cross two edits, so paste the code here.)

This second check can be removed if we refactor the first one after consuming "default" token.

```ts
  const isDefaultExport = consumeOpt(parser, context | Context.AllowRegExp, Token.DefaultKeyword);

  if (parser.leadingDecorators.decorators.length && parser.getToken() !== Token.ClassKeyword) {
    parser.report(Errors.InvalidLeadingDecorator);
  }

  if (isDefaultExport) {
```

No need to check `Token.Decorator` as `@x export @y` is invalid anyway.

test/parser/next/__snapshots__/decorators.ts.snap
```diff
@@ -84,6 +138,12 @@ exports[`Decorators > Decorators (fail) > @foo export default @bar class A {} 1`
     |                     ^ Unexpected token: '@'"
```
3cp: This error will be changed to `Leading decorators must be attached to a class declaration` after my suggested change, I think it's better.

3cp (changes requested):

morgan-coded:
Addressed the round-two review.

- Collapsed both export-side leading-decorator checks into the single post-`default` gate; the `@foo export @bar class A {}` and `@foo export default @bar class A {}` snapshots now read `Leading decorators must be attached to a class declaration`.
- The `Errors.UnexpectedToken, '@'` guard in `parseClassDeclaration` now appears unreachable because all three callers gate a nonempty leading-decorator list on the current token being `class`; would you prefer that guard removed or left as defense-in-depth?

Gates at `89366c9`: 143 files / 91,313 tests, lint, build, and diff-check green.

3cp:
> * The `Errors.UnexpectedToken, '@'` guard in `parseClassDeclaration` now appears unreachable because all three callers gate a nonempty leading-decorator list on the current token being `class`; would you prefer that guard removed or left as defense-in-depth?

Good catch. Yesterday I was thinking about changing your check right after the next parseDecorators, but decided to skip that. But now it looks clearer with the full picture.

The parseClassDeclaration has a blind `nextToken(...)` call designed to consume `class` keyword, which is fine long ago before we added decorator support, as parseClassDeclaration is triggered only when next token is `class`. But now both `@` and `class` can route into parseClassDeclaration, there is no guard to ensure `class` keyword after decorators. I think if we rewrite the code, we can close all loose part, and also make no assumption that leadingDecorators were handled perfectly.


```ts
  if (parser.leadingDecorators.decorators.length) {
    // if (parser.getToken() === Token.Decorator) {
    //   parser.report(Errors.UnexpectedToken, '@');
    // }
    start = parser.leadingDecorators.start!;
    decorators = [...parser.leadingDecorators.decorators];
    parser.leadingDecorators.decorators.length = 0;
  } else {
    start = parser.tokenStart;
    decorators = parseDecorators(parser, context, privateScope);
    // if (decorators.length && parser.getToken() !== Token.ClassKeyword) {
    //   parser.report(Errors.InvalidLeadingDecorator);
    // }
  }

  context = (context | Context.InConstructor | Context.Strict) ^ Context.InConstructor;

  // This handles all cases
  consume(parser, context, Token.ClassKeyword);

  // Could use explicitly Errors.InvalidLeadingDecorator, but I think the above error
  // is clear enough with much less code.

  // nextToken(parser, context);
  //
  // if (parser.getToken() !== Token.ClassKeyword) {
  //   if (decorators.length && parser.getToken() !== Token.ClassKeyword) {
  //     parser.report(Errors.InvalidLeadingDecorator);
  //   } else {
  //     parser.report(Errors.UnexpectedToken, KeywordDescTable[parser.getToken() & Token.Type]);
  //   }
  // }
```

The downside is that in script mode, the error will become `Expected 'class'` which is still very readable.

```
SyntaxError [1:5-1:8]: Expected 'class'
> 1 | @dec var a = 1
    |      ^^^ Expected 'class'
```

3cp:
Oh, the final Errors.InvalidLeadingDecorator check in parseModuleItem can be removed. All three branches in parseModuleItem has the check.

morgan-coded:
Reworked `parseClassDeclaration` around an explicit `class` consume.

- Applied the rewrite as pasted, and removed the final post-switch `parseModuleItem` check.
- Confirmed the predicted snapshot consequence: script-mode failures now report `Expected 'class'`. The same message now covers the `@dec[0] class A {}` restriction family and the feature-off `export default @decorator class Foo {}` case, which previously reported `Expected '{'`.
- One thing the snapshots understate: with decorators off, a leading `@` now reports `Expected 'class'` at the `@` rather than a downstream `Expected '{'`, and the suite covers that path with only the one case above. Everything affected errored before and still errors — in an old-vs-new differential over the decorator and class forms I exercised, nothing changed accept/reject and the ASTs matched wherever parsing succeeded. Happy to leave it as is or narrow it to the feature-on path.

At `523bc7c`: 143 files / 91,313 tests passing; lint, build, and diff-check passing.

3cp:
> a leading @ now reports Expected 'class' at the @ rather than a downstream Expected '{', 

I noticed it. I think the message is not too bad. A fix could be don't route `@` to parseClassDeclaration when decorator is off. But that'll be few more lines of code. Let's ignore it.

Thank you for the work! Will merge soon.

3cp (approved):

3cp:
v7.3.0 is released.

morgan-coded:
Thanks for the invite — I accepted it. I'll keep opening PRs from my fork and leave the merges with you and fisker, same as before. Would you rather I push small fixes straight to `main`? I'm available for review or triage too.

3cp:
You should be able to create branches directly here now. Let's use PR even for small fixes. We appreciate your contributions!

PR #615, opened 2026-07-27
fix(parser): scope pending octal errors to the code unit that produced them

gaoflow:
`5; "\145"; function g() { "use strict"; }` is rejected with *Octal literals are not allowed in strict mode*, pointed at the nested directive. The escape sits in sloppy script code where it is legal — `g` being strict says nothing about its enclosing scope.

The pending-octal flag added for #149 is set by the lexer and read by the next `"use strict"` prologue, but it is never reset at a code-unit boundary, so any nested strict function arms it. Same for `\8`/`\9` and `08`/`09`. This clears the two flags on entry to a function body; a body's own prologue is scanned after that point, so `function f() { "\145"; "use strict"; }` and `"\145"; "use strict";` still fail.

Differential run against acorn and V8 over 615 sources (7 sloppy literals x 12 nested code-unit kinds x 2 nesting depths, plus controls): 253 false rejections before, 0 after, with the two oracles agreeing on all 615. Full test262 at the pinned commit goes 94550 -> 94551 valid programs accepted while 8853 invalid programs are still rejected, so nothing was loosened. Mutating the fix both ways — dropping the reset, moving it after the `{`, resetting only `Flags.Octal`, deleting each prologue check in turn — reddens the new tests every time.

Drops `staging/sm/strict/directive-prologue-01.js` from `test262/whitelist.txt`; that fixture calls these very cases `notAnError1`..`notAnError3`. The line arrived with an automated corpus update (#357) rather than as a deliberate skip. The other nine entries are unrelated causes and stay.


3cp (approved):
Thx! ❤️

3cp:
v7.3.0 is released.

PR #616, opened 2026-07-29
Do we need this? chore: improve error message on "@" when next flag is off

3cp:
I am not convinced whether we need to add these few lines just to make the error message better.

On the other side, not like other stage 3 features, the decorator could stay on stage 3 for a long long time as it has been for the last decade.

cc @morgan-coded

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -259,6 +259,10 @@ function parseStatementListItem(
       );
 
     case Token.Decorator: // @decorator
+      if (!(parser.features & Features.Decorators)) {
+        parser.report(Errors.UnexpectedToken, '@');
+      }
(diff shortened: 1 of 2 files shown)


fisker (approved):
LGTM. Thanks!

3cp:
> LGTM. Thanks!

The question is do we really need these few more lines? :-)

fisker:
> > LGTM. Thanks!
> 
> The question is do we really need these few more lines? :-)

I think so, it's useful.

morgan-coded:
Agreed. The decorators-off error is clearer now, and #617 covers the class-expression path from #612. Thanks for carrying both over.

PR #617, opened 2026-07-29
fix(parser): fix class keyword check after decorator for class expression

3cp:
This follows the same fix in #612.

cc @morgan-coded

### DIFF
--- a/test/parser/next/__snapshots__/decorators.ts.snap
+++ b/test/parser/next/__snapshots__/decorators.ts.snap
@@ -1,15 +1,21 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`Decorators > Decorators (fail) > (@bar const foo = 1); 1`] = `
-"SyntaxError [1:16-1:17]: Expected '{'
+"SyntaxError [1:6-1:11]: Expected 'class'
 > 1 | (@bar const foo = 1);
-    |                 ^ Expected '{'"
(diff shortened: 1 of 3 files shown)


test/parser/next/__snapshots__/decorators.ts.snap
```diff
@@ -1,15 +1,21 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`Decorators > Decorators (fail) > (@bar const foo = 1); 1`] = `
-"SyntaxError [1:16-1:17]: Expected '{'
+"SyntaxError [1:6-1:11]: Expected 'class'
 > 1 | (@bar const foo = 1);
-    |                 ^ Expected '{'"
+    |       ^^^^^ Expected 'class'"
```
fisker, line 6: What will happen when parsing module?

Do we support this?

```
@dec export class {}
```

3cp: Yes there are test cases for leading decorators in module mode.

fisker (approved):

test/parser/next/__snapshots__/decorators.ts.snap
fisker: Ah, message for module parsing

```
[1:5-1:10]: Leading decorators must be attached to a class declaration

> 1 | @bar const foo = 1
    |      ^^^^^
```

PR #618, opened 2026-07-29
chore(ci): add actionlint configuration and workflow

chernojagne:
Adds actionlint to catch GitHub Actions workflow bugs before they reach CI.

**What this adds:**

- `.github/actionlint.yaml` — actionlint configuration with ShellCheck integration enabled. Suppresses four rules that produce common false positives in GitHub Actions `run:` blocks due to `${{ }}` expression interpolation: SC2086, SC2129, SC2155, SC1091.
- `.github/workflows/actionlint.yml` — dedicated workflow that runs actionlint on changes to `.github/workflows/**` and the config file itself.

**Why:**

The existing CI pipeline (eslint, tsc, prettier, cspell, knip) enforces a high-quality code standard. actionlint closes the remaining gap: none of the three current workflows are validated for Actions schema correctness or shell-level bugs in `run:` blocks.

---

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.


3cp:
Thanks! But I am not sure we need this. @fisker can decide on this.

chernojagne:
Thanks for the heads up — happy to wait for @fisker's call, no rush at all.

fisker:
We can add this check, but please add to the existing lint job, take this as example.

https://github.com/prettier/prettier/blob/81e217a234d0114f69e790ce12a436abd07a2e04/.github/workflows/lint.yml#L64

fisker:
As we are here, if you don't mind, we can also add check for renovate config

https://github.com/prettier/prettier/blob/81e217a234d0114f69e790ce12a436abd07a2e04/.github/workflows/lint.yml#L75

chernojagne:
Thanks! Done in 74c3838 — moved the actionlint step into the existing lint job following the prettier/prettier pattern, and removed the standalone workflow file. Also added the renovate config validation step as suggested.

fisker (approved):
Thank you!

Issue #273, opened 2024-07-12
An edge case in `for..in` can't be parsed.


fisker:
Babel recently added a test file

https://github.com/babel/babel/blob/dc1f486ea52bef2341048c499ca1342d5cab7f8a/packages/babel-generator/test/fixtures/parentheses/in-inside-for/input.js#L1

Meriyah was not able to parse this case

```js
for (var a = (x = b in c) => {} in {});
```

Anyway, nobody should really care about this case, but the error location is not correct, maybe worth to look into?

When I use

```js
// https://github.com/babel/babel/blob/HEAD/packages/babel-generator/test/fixtures/parentheses/in-inside-for/input.js

for (var a = (b in c) in {});
for (var a = 1 || (b in c) in {});
for (var a = 1 + (2 || (b in c)) in {});
for (var a = (() => b in c) in {});
for (var a = 1 || (() => b in c) in {});
for (var a = (() => { b in c; }) in {});
for (var a = [b in c] in {});
for (var a = {b: b in c} in {});
for (var a = (x = b in c) => {} in {});
```

The error location is at `[3:140]` which is absolutely wrong

Tcjmd:
Why can't this be parsed? The other examples parse?

3cp:
It's not a valid code.
I think @fisker was saying meriyah generates error on wrong position.

I know where to change the behavior: to allow "in" keyword after an arrow function expression. But it feels like something should be controlled by precedence, so I am not confident about the change in my mind.

Also I am not clear about other side effects of the change, probably need some study of the ecma spec before making a fix.

Tcjmd:
`in` can't have a presedence value. Why not save the bitmask on a temp var before going into next scope and restore it afterwards to original state when done parsing the underlying scope.

För better perf. Make sure to automatically set the bit mask where it should be allowed.

If the case mentioned should be valid the bitmask is off in first part in for stmt, but not set in paren parsing.

Try set the bit in assignment expr and undset it

fisker:
> It's not a valid code.

I think it's valid. But I don't really care if it can be parsed since nobody use it, more care about the error location.

3cp:
Got this very clear now.

When test in browser console.

```
// This is invalid
function test() { 'use strict'; for (var a = (b in c) in {}); }
> SyntaxError: Cannot assign to the loop variable inside a for-in loop header.

// This is valid
function test() { for (var a = (b in c) in {}); }
> undefined
```

[...]

PR #621, opened 2026-07-31
fix for-in loop with initializer

3cp:
The first commit fixed the webcompat check for the initializer.
The second commit fixed the check of 'in' keyword after arrow func expression.

I am not happy with the 2nd change. I used default argument and lots of passing of "origin" in order to just get the `Origin.ForStatement` to parseArrowExpression.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -123,7 +123,7 @@ function parseStatementList(parser: Parser, context: Context, scope: Scope | und
   }
 
   while (parser.getToken() !== Token.EOF) {
-    statements.push(parseStatementListItem(parser, context, scope, undefined, Origin.TopLevel, {}) as ESTree.Statement);
+    statements.push(parseStatementListItem(parser, context, scope, undefined, {}, Origin.TopLevel) as ESTree.Statement);
   }
   return statements;
 }
@@ -199,7 +199,7 @@ function parseModuleItem(parser: Parser, context: Context, scope: Scope | undefi
       moduleItem = parseImportDeclaration(parser, context, scope);
       break;
     default:
-      moduleItem = parseStatementListItem(parser, context, scope, undefined, Origin.TopLevel, {});
+      moduleItem = parseStatementListItem(parser, context, scope, undefined, {}, Origin.TopLevel);
   }
 
   return moduleItem;
@@ -218,8 +218,8 @@ function parseStatementListItem(
   context: Context,
   scope: Scope | undefined,
   privateScope: PrivateScope | undefined,
-  origin: Origin,
   labels: ESTree.Labels,
+  origin: Origin = Origin.BlockStatement,
 ): ESTree.Statement {
   // ECMA 262 10th Edition
   // StatementListItem[Yield, Return] :
@@ -251,11 +251,11 @@ function parseStatementListItem(
         context,
         scope,
         privateScope,
-        origin,
         1,
         HoistedFunctionFlags.None,
         0,
         start,
+        origin,
       );
 
     case Token.Decorator: // @decorator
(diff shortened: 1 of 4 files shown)


3cp:
Fixes #273

src/parser.ts
```diff
@@ -7433,8 +7468,13 @@ function parseArrowFunctionExpression(
         break;
       default: // ignore
     }
-    if ((parser.getToken() & Token.IsBinaryOp) === Token.IsBinaryOp && (parser.flags & Flags.NewLine) === 0)
-      parser.report(Errors.UnexpectedToken, KeywordDescTable[parser.getToken() & Token.Type]);
+    if ((parser.getToken() & Token.IsBinaryOp) === Token.IsBinaryOp && (parser.flags & Flags.NewLine) === 0) {
+      if (parser.getToken() !== Token.InKeyword || origin !== Origin.ForStatement) {
```
3cp, line 7478: The way I passed "origin" down the call chain to here is not pretty.

fisker (approved):

3cp:
I will leave this open for some days in case we found some way to improve the fix.

morgan-coded:
Running `parseScript('for (var a = (x => {} in y);;);', { webcompat: true })` at the PR head accepts the input; V8, Acorn, Babel, and Espree all reject it. The merge base rejects the same input, so this is a regression.

`parseParenthesizedExpression` has already cleared `Context.DisallowIn`, but the block-arrow special case suppresses the error on `Origin.ForStatement` alone; could that check be gated on both the outer for-in initializer state and the active no-`in` context?

The [Annex B for-in initializer grammar](https://tc39.es/ecma262/#sec-initializers-in-forin-statement-heads) also permits `for (var a = async () => {} in b);` and `for (var a = x = () => {} in b);`; both still reject here, but those gaps predate #621.

3cp:
> 'for (var a = (x => {} in y);;);'

I like that, looks like we need to need to reset "origin" the `(` token and also remember previous "origin", pop back previous at `)`.

morgan-coded:
Tested the four-call-site parenthesized-content reset with `Origin.None` inside the group and `Origin.ForStatement` kept on the two arrow-head paths after `)`. That removed the grouped-arrow regression while preserving the Annex B for-in initializer.

The unchanged suite still passed, test262 had zero delta from the PR head, and lint stayed green.

The ambient parser-state version failed as an approach because initializer origin also reaches non-parenthesized nesting, where there is no `(` boundary to reset:

```js
for (var a = {p: () => {} in y};;);
for (var a = f(() => {} in y);;);
for (var a = function () { return () => {} in y; };;);
```

PR #621 and the four-call-site reset rejected all three, as did V8, Acorn, Babel, and Espree; the ambient-state shape accepted all three under every `webcompat`/`lexical` combination. Neither the unchanged suite nor lint caught those accepts. I didn't take the ambient-state shape to test262 once those accepts showed up.

If useful, I can send the four-call-site patch or open a PR against your branch; otherwise I'll leave the fix with you. The pre-existing async-arrow and assignment-arrow gaps stayed unchanged and look separate from this mechanism.

3cp:
Yes please, I am not actively working on this, my second commit is not clean, more like a draft. I appreciate your help, you can ignore my second commit if that helps. If you work on top of the second commit, please clean it up wherever you see fit.

I think your permission would allow you to directly add commits to this branch. No need another PR.

morgan-coded:
Added the four-call-site `Origin.None` reset and regression coverage on top of your branch. The tests cover the grouped-arrow leaks plus stays-rejected object, call, and nested-return cases; I left the rest of the draft unchanged because the remaining edits would alter the mechanism.

All local gates pass: test262 is zero-delta, lint and `git diff --check` are green, and the Node 26/24/22/20 suites pass. It's back in your court for review.

3cp:
Thanks very much! I will have some thought on whether we can clean up the origin passing before merging this.

3cp:
I pushed a commit to refactor all "origin" param.

* put "origin" as last param as possible for all function definitions.
* use default value for the "origin" param, most are Origin.None.
* cleaned up many function calls to omit passing Origin.None, since we have a default value.

morgan-coded:
Verified the refactor at `077e1d5` — no behaviour change in the origin-sensitive surface.

The four parenthesized-content reset sites still resolve to `Origin.None`: three inherit it from the new default, and `parsePrimaryExpression` still passes it explicitly.

Gates at that head: 139 files / 91,540 tests green, `TEST262=1` 142 / 91,543 green. I also built the parser at `a20f19f` and at `077e1d5` and diffed them over 59 origin-sensitive sources — for-in/for-of heads, grouped and async arrows, destructuring patterns, rest elements — across five `webcompat`/`lexical`/`next` combinations in both goals: 590 pairs, zero accept/reject differences, zero AST drift. The Annex B initializer still parses; the grouped-arrow, object, call, and nested-return leaks all still reject.

Nothing outstanding from my side.

3cp:
Really appreciate your diligence!

PR #622, opened 2026-07-31
fix: forbid `}` and `>` in JSX element children

### LINKED ISSUE #619: Forbid `}` in JSX element
This should be invalid

```
<div>}</div>
```

Should reject like [babel](https://ast-explorer.dev/#eNoliTEKgDAQBL8iWwv2Ir7CzrOIMUgkJsEYUSR/9w6rnZ154dBiU5dK+rDxRI3IYlazccyauVvs1ZeukWEVWL3kq4qQQj60GZ5oCC3/PSzZMdd/ji6v1idpoxhxW7oJcibyBeUDydEnSA==), [espree(acorn)](https://ast-explorer.dev/#eNpNjTEKwzAMRa8SNAe6h9CxJyidvBjnDw6ObSQ7tATfvXJDSyfpvyfxDwo00Wp3K459LjRSVgDJDGhwGubF79c2X/pQlBQdJg6DIUmVHe6vDEOT5i0tNeg+nhpusw+w+BRPH2yBlJ8PyXVeuOLv4wZbKkO6+vR0scrze9pRM7FRewMbwjwe)

```
Unexpected token `}`. Did you mean `&rbrace;` or `{'}'}`? (1:5)
```

a1stok:
fixes #619

this fixes an issue where a stray `}` in JSX text (like `<div>}</div>`) was silently ignored instead of throwing an error. 

the problem was that `}` wasn't marked as a JSX token boundary in the character classifier, so the lexer just ate it up as normal JSX text. this adds the correct flag for `}` so the lexer stops at it and correctly throws an unexpected token error like babel and espree do. also fixed a small issue where `parseJSXText` was calling the normal `nextToken` instead of `nextJSXToken` which was causing the wrong error message to show up.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -9208,7 +9208,6 @@ function parseJSXChildOrClosingFragment(
  * Parses JSX Text
  *
  * @param parser Parser object
- * @param context  Context masks
  */
 function parseJSXText(parser: Parser): ESTree.JSXText {
   const start = parser.tokenStart;
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -9166,7 +9166,7 @@ function parseJSXChildOrClosingElement(
   privateScope: PrivateScope | undefined,
   inJSXChild: 0 | 1,
 ) {
-  if (parser.getToken() === Token.JSXText) return parseJSXText(parser, context);
+  if (parser.getToken() === Token.JSXText) return parseJSXText(parser);
   if (parser.getToken() === Token.LeftBrace)
     return parseJSXExpressionContainer(parser, context, privateScope, /*inJSXChild*/ 1, /* isAttr */ 0);
   if (parser.getToken() === Token.LessThan) {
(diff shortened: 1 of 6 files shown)


3cp (approved):
Thank you!

3cp:
Pls run npx vitest -u to update the error message in snapshot file.

a1stok:
sure will do

fisker:
`>` should be forbidden as well.

fisker:
https://github.com/acornjs/acorn-jsx/blob/8ed96d6ddec2065204ba07d924bb2e7bca539ea6/index.js#L139

https://github.com/babel/babel/blob/06b6eae39da4ce0689fad64e2c48a6375a464208/packages/babel-parser/src/plugins/jsx/index.ts#L139

fisker (approved):

fisker:
@a1stok Lint is failing.

src/parser.ts
```diff
@@ -9210,7 +9210,7 @@ function parseJSXChildOrClosingFragment(
  * @param parser Parser object
  * @param context  Context masks
```
fisker:
```suggestion
```

3cp:
@fisker is it ready for merge? Change commit message to "fix(jsx):..." when merging, thx!

fisker (approved):

PR #623, opened 2026-07-31
fix: forbid await in parenthesized async arrow parameters

### LINKED ISSUE #337: async () problem

### LINKED ISSUE #620: Should not allow `await` in async function parameters
This should be invalid.

```
async (x = await (2)) => {};
```

a1stok:
fixes #620

this fixes a bug where `await` was incorrectly allowed inside parenthesized default parameters of async arrow functions (e.g. `async(x = await (0)) => 1`).

the parser was correctly checking for `await` in normal parameters, but when it hit a parenthesized expression it would create a new parsing scope and accidentally wipe the await flag. this just saves the await state before parsing the parens and restores it right after, so we don't lose track of the fact that we're still in a parameter list that contains an await.

Closes #337


3cp (approved):
Thank you!

3cp:
Some lint error. You can do npx prettier . --write

test/parser/expressions/__snapshots__/async-arrow.ts.snap
```diff
@@ -1368,6 +1368,12 @@ exports[`Expressions - Async arrow > Expressions - Async arrow (fail) > async (x
     |                   ^^^^^^^^^^^^^ Illegal 'use strict' directive in function with non-simple parameter list"
 `;
 
+exports[`Expressions - Async arrow > Expressions - Async arrow (fail) > async (x = await (2)) => {}; 1`] = `
+"SyntaxError [1:22-1:24]: Await expression not allowed in formal parameter
+> 1 | async (x = await (2)) => {};
+    |                       ^^ Await expression not allowed in formal parameter"
```
fisker: Can this location be improved?

a1stok: made it highlight the `await` word instead of `=>`, or was this about something else?

3cp:
```
[1:11-1:18]: Unexpected strict mode reserved word

> 1 | function a(package) { 'use strict'; }
    |            ^^^^^^^
```

We have some existing practice that remembers the previous token location for a later error.

We could do similar thing. But if it's too troublesome, I don't mind to accept the PR in current state.

a1stok: yep, I did add it, added `firstAwaitLocation `to track the position of the first `await ` token, so it directly points at it

3cp: I am confused. Why the snapshot error did not point to the "await" token?

3cp: @a1stok, in script mode, you got the correct location, but in module mode, you missed.

script mode:
```
[1:11-1:16]: Await expression not allowed in formal parameter

> 1 | async (x = await (2)) => {};
    |            ^^^^^
```

module mode:
```
[1:22-1:24]: Await expression not allowed in formal parameter

> 1 | async (x = await (2)) => {};
    |                       ^^
```

a1stok:
@fisker any other fixes needed for  this one?

3cp (approved):

test/parser/expressions/__snapshots__/async-arrow.ts.snap
3cp: @a1stok the miss in module mode is still not fixed.

3cp (changes requested):

test/parser/expressions/__snapshots__/async-arrow.ts.snap
a1stok: @3cp didn`t see the comment somehow, just fixed it

3cp (approved):
Thank you!

PR #627, opened 2026-08-02
fix: allow top-level `using` declarations in commonjs source type

### LINKED ISSUE #626: Support top level `using` in commonjs
```
> require('meriyah').parse('using foo = bar', {sourceType:'commonjs'})
Uncaught [ParseError [SyntaxError]: [1:6-1:9]: Unexpected token: 'identifier'
] {
  start: 6,
  end: 9,
  range: [ 6, 9 ],
  loc: { start: { line: 1, column: 6 }, end: { line: 1, column: 9 } },
  description: "Unexpected token: 'identifier'"
}
> require('meriyah').parse('using foo = bar', {sourceType:'module'})
{
  type: 'Program',
  sourceType: 'module',
  body: [
    {
      type: 'VariableDeclaration',
      kind: 'using',
      declarations: [Array]
    }
  ]
}
```

~Should be allowed in 'commonjs', forbidden in 'module'.~ Should both be allowed.

morgan-coded:
Fixes #626

Node accepts a top-level `using` declaration in a `.cjs` wrapper, while `vm.Script` rejects the same declaration under the Script goal; test262's early error permits `UsingDeclaration` inside a `FunctionBody`.

Meriyah already models CommonJS top level as a function body for `return` and `new.target`, so this exempts that context from the top-level Script guard.

Script goal still rejects top-level `using`, and CommonJS still rejects top-level `await using`; both have new negative coverage.

The case was already parked in `test/parser/miscellaneous/commonjs.ts` from #524 as the third row of JLHwung's acorn table after `return` and `new.target`.

I ran the 139-file suite (91,499 tests), the 142-file test262 suite (91,502 tests), production tests, build, lint, and conformance under Node 20, 22, 24, and 26.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1744,7 +1744,12 @@ function parseUsingDeclarationOrExpressionStatement(
   const expression = parseIdentifier(parser, context);
 
   if ((parser.flags & Flags.NewLine) === 0 && isResourceBindingStart(parser.getToken())) {
-    if (origin & Origin.TopLevel && context & Context.InGlobal && (context & Context.Module) === 0) {
+    if (
+      origin & Origin.TopLevel &&
+      context & Context.InGlobal &&
+      (context & Context.Module) === 0 &&
+      (context & Context.InReturnContext) === 0
+    ) {
       parser.report(Errors.UnexpectedToken, KeywordDescTable[parser.getToken() & Token.Type]);
     }
 
--- a/test/parser/miscellaneous/__snapshots__/commonjs.ts.snap
+++ b/test/parser/miscellaneous/__snapshots__/commonjs.ts.snap
@@ -18,12 +18,6 @@ exports[`Statements - Return > Commonjs (fail) > return 1`] = `
     | ^^^^^^ Illegal return statement"
 `;
 
-exports[`Statements - Return > Commonjs (fail) > using foo = null 1`] = `
-"SyntaxError [1:6-1:9]: Unexpected token: 'identifier'
-> 1 | using foo = null
-    |       ^^^ Unexpected token: 'identifier'"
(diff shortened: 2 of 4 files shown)


src/parser.ts
```diff
@@ -1744,7 +1744,12 @@ function parseUsingDeclarationOrExpressionStatement(
   const expression = parseIdentifier(parser, context);
 
   if ((parser.flags & Flags.NewLine) === 0 && isResourceBindingStart(parser.getToken())) {
-    if (origin & Origin.TopLevel && context & Context.InGlobal && (context & Context.Module) === 0) {
+    if (
+      origin & Origin.TopLevel &&
+      context & Context.InGlobal &&
+      (context & Context.Module) === 0 &&
+      (context & Context.InReturnContext) === 0
+    ) {
```
fisker, line 1752: Should we add a new flag here? https://github.com/meriyah/meriyah/blob/78f9544fc755e156c9dfb24e3b419684f721cd01/src/parser.ts#L53

3cp: ~~I think no. Commonjs only set initial context to InReturnContext, that context can change to off for example in class body.~~

3cp: Sorry, you mean add a context for global using.

3cp (approved):

src/parser.ts
3cp: If add another flag, then we need to set that flags properly at start of all kinds of block to enable positive cases for using declaration.

Right now, there are only two negative cases: (1) top level, (2) case/default branches in switch statement. They are easy to handle.

Let's keep this implementation.

fisker (approved):

PR #628, opened 2026-08-12
fix(parser): fix error location for duplicate binding

3cp:
The var/block name was added in scope without location information. Most of the time, the parser has moved to next token when reporting duplicate binding.

This fix retains location (start and end) information when var/block name is added.

The difference in snap files shows what this fixed.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -2789,6 +2789,7 @@ function parseImportSpecifierOrNamedImports(
 
   while (parser.getToken() & Token.IsIdentifier || parser.getToken() === Token.StringLiteral) {
     let { tokenValue, tokenStart, currentLocation } = parser;
+    const start = tokenStart;
     const token = parser.getToken();
     const imported = parseModuleExportName(parser, context);
     let local: ESTree.Identifier;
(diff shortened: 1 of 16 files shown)


fisker (approved):
👍

morgan-coded (commented):
Compared the new locations against acorn and babel: 21/21 agreement across lexical/var/function/class collisions, strict and destructured parameters, catch, for-heads, imports, and exports, and 10 valid shadowing controls still parse with byte-identical ASTs. The range threading stays parameter-scoped and reads clean.

One edge diverges: deferred parameter duplicates with three or more repeats. For `(a, a, a)` this branch reports the third `a` (offset 7) where acorn and babel report the second (offset 4) — the first occurrence that makes the binding set invalid. `recordScopeError` assigns `this.scopeError` unconditionally, so each later repeat overwrites the captured range; two repeats agree with the peers everywhere, and the same divergence shows for strict functions, async arrows, and destructured parameters, growing with a fourth repeat. Preserving the first recorded duplicate range would line these up. Happy to leave it as a follow-up if you'd rather land this as is — acceptance is unchanged either way, and the full suite plus test262 are green on this head locally.

3cp:
> `recordScopeError` assigns `this.scopeError` unconditionally

Probably can be solved by assigning `this.scopeError` only when it's absent. Let me try.

3cp:
Unbelievable :)

I started with "//", trying to add a comment then add implementation. GitHub Copilot automatically suggested the comment (with no hint!) and then automatically suggested the new code. Note this is my local Copilot in vscode, I don't think it read the comment on the PR.

```js
recordScopeError(type: Errors, tokenStart: Location, tokenEnd: Location, ...params: string[]) {
    // Only record the first error, subsequent errors will be ignored
    if (this.scopeError) {
      return;
    }
    this.scopeError = {
```

fisker:
> Unbelievable :)

We are doomed.

----

I suggest

```
this.scopeError ??= ...
```

PR #630, opened 2026-08-16
fix(jsx): fix pasring jsx contextual identifier

### LINKED ISSUE #629: Allow attributes with `-` in JSX
I didn't check the spec, but it looks like all major parsers allow `-` in JSX attributes.

```jsx
<a with-dash={1} />
```

3cp:
Contextual jsx identifier like "with" which is a normal JS reserved word.

close #629

### DIFF
--- a/test/parser/miscellaneous/jsx.ts
+++ b/test/parser/miscellaneous/jsx.ts
@@ -525,5 +525,6 @@ describe('Miscellaneous - JSX', () => {
     { code: '<a></* block */\n/a>;', options: { jsx: true } },
     { code: '</* open fragment */>\n</ /* close fragment */>;', options: { jsx: true, ranges: true } },
     { code: '<a><  /a>', options: { jsx: true, ranges: true } },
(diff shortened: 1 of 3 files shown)


test/parser/miscellaneous/jsx.ts
```diff
@@ -525,5 +525,6 @@ describe('Miscellaneous - JSX', () => {
     { code: '<a></* block */\n/a>;', options: { jsx: true } },
     { code: '</* open fragment */>\n</ /* close fragment */>;', options: { jsx: true, ranges: true } },
     { code: '<a><  /a>', options: { jsx: true, ranges: true } },
+    { code: '<a with-dash={1} />', options: { jsx: true, ranges: true } },
```
fisker: More tests and drop `ranges`?

```suggestion
    { code: '<a with-dash />', options: { jsx: true } },
    { code: '<with-dash />', options: { jsx: true } },
    { code: '<a with />', options: { jsx: true } },
    { code: '<with/>', options: { jsx: true } },
````

fisker (approved):

PR #632, opened 2026-08-16
fix(jsx): fix tokenizing ">" when "=" is after

### LINKED ISSUE #631: Can't parse `<div>=</div>`
Found this when updating yuku-parser https://github.com/yuku-toolchain/yuku/issues/153

3cp:
close #631

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -28,6 +28,7 @@ export const enum Context {
   DisallowIn = 1 << 17,
   AllowEscapedKeyword = 1 << 18,
   InStaticBlock = 1 << 19,
+  InJSXTag = 1 << 20,
 }
 
 /**
--- a/src/lexer/scan.ts
+++ b/src/lexer/scan.ts
@@ -396,7 +396,7 @@ export function scanSingleToken(parser: Parser, context: Context, state: LexerSt
           return Token.Subtract;
         }
 
-        // `/`, `/=`, `/>`, '/*..*/'
+        // `/`, `/=`, '/*..*/'
         case Token.Divide: {
           advanceChar(parser);
           if (parser.index < parser.end) {
@@ -466,6 +466,10 @@ export function scanSingleToken(parser: Parser, context: Context, state: LexerSt
         // `>`, `>=`, `>>`, `>>>`, `>>=`, `>>>=`
         case Token.GreaterThan: {
           advanceChar(parser);
+
+          if (context & Context.InJSXTag) {
+            return Token.GreaterThan;
+          }
 
           const ch = parser.currentChar;
 
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -9335,7 +9335,7 @@ function parseJSXOpeningElementOrSelfCloseElement(
   const attributes = parseJSXAttributes(parser, context, privateScope);
   const selfClosing = parser.getToken() === Token.Divide;
 
-  if (selfClosing) consume(parser, context, Token.Divide);
+  if (selfClosing) consume(parser, context | Context.InJSXTag, Token.Divide);
 
   if (parser.getToken() !== Token.GreaterThan) {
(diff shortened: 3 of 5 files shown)


src/lexer/scan.ts
```diff
@@ -396,7 +396,7 @@ export function scanSingleToken(parser: Parser, context: Context, state: LexerSt
           return Token.Subtract;
         }
 
-        // `/`, `/=`, `/>`, '/*..*/'
```
3cp, line 399: `/>` was never scanned as a single token. Removed this confusing comment.

src/lexer/scan.ts
```diff
@@ -467,6 +467,10 @@ export function scanSingleToken(parser: Parser, context: Context, state: LexerSt
         case Token.GreaterThan: {
           advanceChar(parser);
 
+          if (context & Context.InJSXTag) {
+            return Token.GreaterThan;
+          }
```
3cp, line 472: The new context `InJSXTag` guards against parsing `>=` (and also `>>`, `>>>`, `>>=`, `>>>=`) in JSX.

src/parser.ts
```diff
@@ -9335,7 +9335,7 @@ function parseJSXOpeningElementOrSelfCloseElement(
   const attributes = parseJSXAttributes(parser, context, privateScope);
   const selfClosing = parser.getToken() === Token.Divide;
 
-  if (selfClosing) consume(parser, context, Token.Divide);
+  if (selfClosing) consume(parser, context | Context.InJSXTag, Token.Divide);
```
3cp, line 9338: For case of `/>=`.

src/parser.ts
```diff
@@ -9639,7 +9639,7 @@ function parseJSXIdentifier(parser: Parser, context: Context): ESTree.JSXIdentif
     parser.report(Errors.UnexpectedToken, KeywordDescTable[parser.getToken() & Token.Type]);
   }
   const { tokenValue } = parser;
-  nextToken(parser, context);
+  nextToken(parser, context | Context.InJSXTag);
```
3cp, line 9642: For case of `>=`.

test/parser/miscellaneous/__snapshots__/jsx.ts.snap
```diff
@@ -205,15 +205,15 @@ exports[`Miscellaneous - JSX > Miscellaneous - JSX (fail) > <a><</a> 1`] = `
 `;
 
 exports[`Miscellaneous - JSX > Miscellaneous - JSX (fail) > <a>></a> 1`] = `
-"SyntaxError [1:2-1:4]: Unexpected token: '>>'
+"SyntaxError [1:3-1:3]: Unexpected token \`>\`. Did you mean \`&gt;\` or \`{'>'}\`?
```
3cp, line 208: The new error is better than before. As we avoided parsing `>>` as single token.

src/lexer/scan.ts
3cp: I recall it was refactored to cater `/[space]>`.

src/common.ts
```diff
@@ -28,6 +28,7 @@ export const enum Context {
   DisallowIn = 1 << 17,
   AllowEscapedKeyword = 1 << 18,
   InStaticBlock = 1 << 19,
+  InJSXTag = 1 << 20,
```
fisker, line 31: `InJSXTag` or `InJsxTag`?

fisker (approved):
Looks great, thank you! Just one nit commit https://github.com/meriyah/meriyah/pull/632#discussion_r3793186225 .

src/common.ts
3cp: Let's use "JSX". All our other places use all-capitals like the "JSXToken".

3cp:
Note we have an `inJSXChild` flag as local variable in some places. Might consider to move it to a context flag in future.

Issue #306, opened 2024-08-06
Deal with "async of" in for and for-await loop.


3cp:
In ECMA spec:
ecma 14.7.5

for ( [lookahead ∉ { let, async of }] [LeftHandSideExpression](https://tc39.es/ecma262/#prod-LeftHandSideExpression)[?Yield, ?Await] of 
  [AssignmentExpression](https://tc39.es/ecma262/#prod-AssignmentExpression)[+In, ?Yield, ?Await] ) [Statement](https://tc39.es/ecma262/#prod-Statement)[?Yield, ?Await, ?Return]

[+Await] for await ( [lookahead ≠ let] [LeftHandSideExpression](https://tc39.es/ecma262/#prod-LeftHandSideExpression)[?Yield, ?Await] of 
[AssignmentExpression](https://tc39.es/ecma262/#prod-AssignmentExpression)[+In, ?Yield, ?Await] ) [Statement](https://tc39.es/ecma262/#prod-Statement)[?Yield, ?Await, ?Return]

In human readable form
```js
for (async of list) {} // Invalid code
for ((async) of list) {} // But this is valid
for await (async of list) {} // Valid code
```

Meriyah has no code dealing with "async of" right now. Need to add it in.

3cp:
```js
for (async of => {}; i < 10; i++); // Valid code
for (\u0061sync of list); // Escaped async is valid
```

kenbern:
@3cp Inspired by Meriyah's challenges that you seem to face all the time I'm converting proprietary code to TypeScript for readability. 
This work-in-progress (4000+ lines) demonstrates:

 * Lexerless parsing
 * High performance
 * Symbol-based scoping
 * Prettier-like formatting
 * ESLint-compatible linting
 * Minification and constant folding
All done in one pass, AST-free, with incremental parsing support. Due to limited time, it's not fully refined.
A few days are needed to create a working example.

 Remaining work includes:

 * Plugin system for linting/formatting
 * Advanced dead code removal
 
While C or assembly would be ideal, I'm confident this code, once optimized, will outperform Meriyah by roughly 80%.

kenbern:
@3cp @fisker Here is simplified demo code to demonstrate alternative parser strategy's. 

https://github.com/kenbern/esquick

kenbern:
@3cp off topic.  I just finished off and completed what you can read about in kink below. 

Let me know if I'm qualified enough to help out with Meriyah

[...]

PR #633, opened 2026-08-25
fix(parser): allow `async of` in for await-of statement

abhijeet117:
## Summary

`for await (async of x)` failed with "Expected '=>'". The for-await head went through the generic expression path where `parseAsyncExpression` treats `async` followed by an identifier like `of` as the start of an async arrow function. The `[+Await]` for-in/of grammar only excludes `let`, so the identifier `async` is valid there (test262 for-await-of/head-lhs-async.js). The for statement now parses `async` in a for-await head as an identifier, mirroring `let` and `using` handling.

## Testing

Reproduced on master: `for await (async of [7])` threw "Expected '=>'" at module top level and inside async functions; both parse now. Added regression tests in for-await-of.ts including rejection of plain `for (async of x)`, plus acceptance cases in for-of.ts and for.ts. Full suite passes (91631 tests); eslint, prettier, tsc, cspell, knip clean.

## Checklist

- [x] Bug reproduced before fix
- [x] Root cause identified
- [x] Bug fixed
- [x] Tests passed


3cp:
#306

Thx! I will test this.

3cp:
The normal `npm test` skips test262 suite.

Our compliance has been improved. Just need to remove two lines in `test262/whitelist.txt`

```
language/statements/for-await-of/head-lhs-async.js (default)
language/statements/for-await-of/head-lhs-async.js (strict mode)
```

To verify, run tests with an env variable: `env TEST262=1 npm test` to trigger full test262 test suite. That's what was checked by CI.

3cp (approved):

abhijeet117:
The Node.js CI failures came from the test262 whitelist check: language/statements/for-await-of/head-lhs-async.js (default and strict mode) now parses successfully as intended by this fix, so its two whitelist entries became violations. The second commit removes them, and I verified locally that the full test262 run passes with zero whitelist violations.

fisker (approved):

Issue #90, opened 2020-07-23
Line/column location is wrong with JSX


moroshko:
Trying to parse the following code using `loc: true, jsx: true`:
```
function App() {
  return (
    <Container>
      <Text>Hello</Text>
    </Container>
  );
}
```
gives an unexpected location for `<Text>`:
```
"loc": {
  "start": {
    "line": 3,
    "column": 23
  },
  "end": {
    "line": 3,
    "column": 27
  }
}
```
I expect it to start on line 4.

Am I missing something?

KFlash:
Obviously an bug! 

cc/ @3cp Do you have any idea for an quick solution?

3cp:
Not just `Text`, `Container` closing tag is wrong too.
It looks like the jsx parser missed all the line feeds to reset line and column. Do you know where is the problematic lines?

KFlash:
@3cp I think it happened around [here](https://github.com/meriyah/meriyah/blob/5fa94bde361167790500cde8bf47fe2eb30271ac/src/lexer/jsx.ts#L52)

3cp:
It looks like scanJSXToken is called twice to scan the same source location (index 3).

The duplicated scan is probably a bug.

The first scan picked up two new lines correctly, the second scan missed the first new line, it strangely thinks the char at index 3 is a `<` instead of `\n`. I guess this comes as a side effect of the duplicated scan.

```ts
export function scanJSXToken(parser: ParserState): Token {
  parser.startPos = parser.tokenPos = parser.index;
  parser.startColumn = parser.colPos = parser.column;
  parser.startLine = parser.linePos = parser.line;

  if (parser.index >= parser.end) return (parser.token = Token.EOF);

  const token = TokenLookup[parser.source.charCodeAt(parser.index)];
  console.log('index: ' + parser.index);
  switch (token) {
    // '<'
    case Token.LessThan: {
      advanceChar(parser);
      if (parser.currentChar === Chars.Slash) {
        advanceChar(parser);
        parser.token = Token.JSXClose;
      } else {
        parser.token = Token.LessThan;
      }

      break;
    }
    // '{'
    case Token.LeftBrace: {
      advanceChar(parser);
      parser.token = Token.LeftBrace;
      break;
    }
    default:
      let state = LexerState.None;

      while (parser.index < parser.end) {
        const type = CharTypes[parser.currentChar];
        console.log('char ' + JSON.stringify(String.fromCharCode(parser.currentChar)));

[...]

PR #635, opened 2026-08-30
fix(jsx): count line terminators and decode entities in attribute values

morgan-coded:
`scanJSXString` left `parser.line` unchanged when a JSX attribute value contained a line terminator, so the literal and every following node kept the wrong location — the same failure shape reported in #90. It also returned raw entity text instead of the strict decode already used for JSX text, following the #133 trail.

This counts LF, CR, CRLF, LS, and PS with the line-tracking idioms already used by `scanString`, then decodes the source slice with `decodeHTMLStrict` as `nextJSXToken` does. `Literal.raw`, byte ranges, and program acceptance stay unchanged; two error positions move to the correct line while their type and message stay the same.

The JSX alignment harness built an acorn-jsx parser but called plain acorn, so it compared 1 of 47 cases and skipped 46 before Meriyah ran. It now compares 40 with an exact-count assertion, while the test262 alignment harness adds a 40,000-fixture floor alongside its retained 90% comparison floor. The harness repair and parser fixes stay together because arming the harness alone makes cases `0010`, `0013`, and `0035` fail.

The focused oracle checks move 5 of 5 terminator forms and 12 of 12 valid entities into agreement with both acorn-jsx and Babel. Across Node 20/22/24/26, the plain suites pass with 91,757/91,757/91,759/91,759 tests, TEST262 with 91,760/91,760/91,762/91,762, and CI with 91,766/91,766/91,768/91,768; `npm run lint`, `npm run lint:snapshots`, `npm run build`, and `npm run production-test` also pass. The test262 whitelist is untouched.

On Node 22, the `dist/meriyah.mjs` parse benchmark was 8.95% slower at the median on a 1.62 MB attribute-heavy JSX corpus, with head slower in 7 of 7 alternating sweeps and a +6.17% to +26.20% range. The 47-case general JSX leg crossed zero (-10.22% to +9.27%), so I do not treat it as a stable directional result.

`0044-16f0.jsx` remains intentionally unaligned because Meriyah uses the full WHATWG entity table chosen in #133 while acorn-jsx leaves `&rbrace;` raw. The exact 40-case assertion is a hand-maintained ratchet over the version-pinned `jsx-test-suite`, in the same shape as #578.

### DIFF
--- a/test/conformance/jsx-ast-alignment-test.ts
+++ b/test/conformance/jsx-ast-alignment-test.ts
@@ -197,7 +197,7 @@ function fixAcornAst(ast: acorn.Program, text: string): MeriyahAst {
         return node;
       case 'JSXOpeningFragment':
         // Not in ESTree, `acorn-jsx` adds them, Babel doesn't
-        // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/estree-jsx/index.d.ts
+        // https://github.com/react/jsx/blob/main/AST.md#jsx-fragment
         delete node.attributes;
         delete node.selfClosing;
         return node;
--- a/test/conformance/jsx-ast-alignment-test.ts
+++ b/test/conformance/jsx-ast-alignment-test.ts
@@ -9,7 +9,22 @@ import { visitNode } from '../test-utils.ts';
 
 const { TEST_JSX_FILE } = process.env;
 
-const notAlignedTests: Set<string> = new Set([]);
+const notAlignedTests: Set<string> = new Set([
+  // `<A>foo&rbrace;</A>`. Meriyah decodes HTML entities in JSX text with the full
+  // https://html.spec.whatwg.org/entities.json table, `acorn-jsx` uses a smaller
+  // one, so entities that only exist in the full table (`&rbrace;`, `&AMP;`, ...)
+  // stay raw in Acorn and are decoded here. That is deliberate, see
+  // https://github.com/meriyah/meriyah/issues/133#issuecomment-770512949
+  '0044-16f0.jsx',
+]);
+
+// How many of the `jsx-test-suite` cases are actually compared against Acorn.
+// `runTest` silently skips every case Acorn cannot parse, so a `parseAcorn` that
+// lost its JSX support skips all of them and this test passes while comparing
+// nothing. `jsx-test-suite` is version pinned, so this is an exact number.
+const EXPECTED_COMPARED_COUNT = 40;
+
+let comparedCount = 0;
 
 it(
   'AST alignment with Acorn (JSX)',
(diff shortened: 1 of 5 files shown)


test/conformance/jsx-ast-alignment-test.ts
```diff
@@ -86,12 +112,12 @@ function parseMeriyah(text: string) {
...
 
 type AcornAst = acorn.Program & { comments: acorn.Comment[] };
-let acornParser;
+let acornParser: typeof acorn.Parser | undefined;
 function parseAcorn(text: string) {
   acornParser ??= acorn.Parser.extend(acornJsx());
 
   const comments: acorn.Comment[] = [];
-  const ast = acorn.parse(text, {
+  const ast = acornParser.parse(text, {
```
fisker, line 120: Can't believe this mistake, thank you!

test/conformance/jsx-ast-alignment-test.ts
```diff
@@ -135,10 +161,46 @@ function fixAcornAst(ast: acorn.Program, text: string): MeriyahAst {
...
+      case 'AccessorProperty':
+      case 'PropertyDefinition':
+      case 'MethodDefinition':
+        if (!('decorators' in node)) {
+          node.decorators = [];
+        }
+        return node;
+      case 'JSXOpeningFragment':
+        // Not in ESTree, `acorn-jsx` adds them, Babel doesn't
+        // https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/estree-jsx/index.d.ts
```
fisker:
```suggestion
        // https://github.com/react/jsx/blob/main/AST.md#jsx-fragment
```

test/conformance/jsx-ast-alignment-test.ts
```diff
@@ -9,7 +9,22 @@ import { visitNode } from '../test-utils.ts';
...
 const { TEST_JSX_FILE } = process.env;
 
-const notAlignedTests: Set<string> = new Set([]);
+const notAlignedTests: Set<string> = new Set([
+  // `<A>foo&rbrace;</A>`. Meriyah decodes HTML entities in JSX text with the full
+  // https://html.spec.whatwg.org/entities.json table, `acorn-jsx` uses a smaller
+  // one, so entities that only exist in the full table (`&rbrace;`, `&AMP;`, ...)
+  // stay raw in Acorn and are decoded here. That is deliberate, see
+  // https://github.com/meriyah/meriyah/issues/133#issuecomment-770512949
+  '0044-16f0.jsx',
```
fisker, line 18: I think it's bug, opened https://github.com/meriyah/meriyah/issues/636 to track.

3cp:
@fisker please merge this when you think it's ready. Thank both of you!

fisker (approved):

PR #637, opened 2026-08-31
fix(parser): allow constructor/accessor identifiers and using arrows

morgan-coded:
Finishes #377 for the two contextual keywords it did not retag: `constructor` and `accessor` now carry `IsExpressionStart | IsIdentifier`, matching `get`, `set`, `from`, and `of`. Without those flags, `await accessor` and `await constructor` collapsed to identifier-`await` and valid module and async forms errored under default parser options; `accessor` later inherited the same gap when auto-accessor support copied the `constructor` token shape. Four unique current `monaco-editor` files fail on published code such as `await accessor.get(...)` today, and all four parse with this change.

The post-fix contextual-keyword check left one residual divergence: bare `using => 0` and `using => {}` still failed at statement position. The statement path now uses the same arrow continuation as `let`, while the other `using` arrow positions and newline rule stay unchanged. `using` came from my #579 work, so this also locks a bug I found in that implementation. `await using => 0` deliberately stays a SyntaxError with `fail()` coverage — V8 and acorn reject it because an arrow used as an `await` operand needs parentheses. Setting `IsIdentifier` also makes `using constructor = x` and `using accessor = x` resource declarations; V8 and Babel accept both bindings.

The base-vs-fixed blast-radius probe covered all 43 `IsExpressionStart` / `IsIdentifier` read sites across 50 families: only eight intended flag deltas moved, with no Node-rejected acceptance. The plain, test262, and CI suites passed across Node 20, 22, 24, and 26 with up to 91,760 tests, along with lint, snapshot lint, build, and production tests; the test262 whitelist was unchanged.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -1736,9 +1736,9 @@ function parseUsingDeclarationOrExpressionStatement(
   labels: ESTree.Labels,
   origin: Origin,
 ): ESTree.VariableDeclaration | ESTree.LabeledStatement | ESTree.ExpressionStatement {
-  const { tokenStart, tokenValue } = parser;
+  const { tokenStart, tokenValue, currentLocation } = parser;
   const token = parser.getToken();
-  const expression = parseIdentifier(parser, context);
+  let expression: ESTree.Identifier | ESTree.Expression = parseIdentifier(parser, context);
 
   if ((parser.flags & Flags.NewLine) === 0 && isResourceBindingStart(parser.getToken())) {
     if (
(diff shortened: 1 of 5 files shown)


3cp (approved):

src/parser.ts
```diff
@@ -1764,6 +1764,28 @@ function parseUsingDeclarationOrExpressionStatement(
...
+
+  if (parser.getToken() === Token.Arrow) {
+    let arrowScope: Scope | undefined = void 0;
+
+    if (parser.options.lexical)
+      arrowScope = createArrowHeadParsingScope(parser, context, tokenValue, tokenStart, currentLocation);
+
+    parser.flags = (parser.flags | Flags.NonSimpleParameterList) ^ Flags.NonSimpleParameterList;
+
+    expression = parseArrowFunctionExpression(
```
3cp: Looks like the type change on line 1741 is unnecessary. The `ESTree.Expression` is only needed here. You can create another local var on line 1776 instead.

The type change on line 1741 confuses reader.

3cp (changes requested):

3cp (approved):

PR #638, opened 2026-09-01
fix(jsx): fix named entries decoding in `JSXText`

### LINKED ISSUE #636: Use named character references list from HTML 4
The JSX spec explicitly defines which named entries should be supported.

```
[HTMLNamedCharacterReferenceName](https://react.github.io/jsx/#prod-HTMLNamedCharacterReferenceName) ::
List of 252 character entity references defined in HTML4 standard
```

> This means that the later expansion of this list to [2231 named character references](https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references) since HTML5 are intentionally excluded from the support.

> The list of 252 character entity references defined in HTML4 standard can be found [at here](https://www.w3.org/TR/1999/REC-html401-19991224/sgml/entities.html).

https://react.github.io/jsx/#sec-HTMLCharacterReference

fisker:
Fixes #636

Follow JSX spec, only decode HTML4 entries.


test/conformance/jsx-ast-alignment-test.ts
```diff
@@ -9,14 +9,7 @@ import { visitNode } from '../test-utils.ts';
...
 
-const notAlignedTests: Set<string> = new Set([
-  // `<A>foo&rbrace;</A>`. Meriyah decodes HTML entities in JSX text with the full
-  // https://html.spec.whatwg.org/entities.json table, `acorn-jsx` uses a smaller
-  // one, so entities that only exist in the full table (`&rbrace;`, `&AMP;`, ...)
-  // stay raw in Acorn and are decoded here. That is deliberate, see
-  // https://github.com/meriyah/meriyah/issues/133#issuecomment-770512949
-  '0044-16f0.jsx',
-]);
+const notAlignedTests: Set<string> = new Set([]);
```
3cp, line 12: Do we still need this?

fisker: We don't, but keeping this makes it easier to skip tests in the future.

3cp (approved):

PR #642, opened 2026-09-01
fix(jsx): fix `JSXText` value decoding

### LINKED ISSUE #641: More bugs in JSXText value decoding
1. Should only decode hex characters with lowercase `x` not `&#X41;` https://react.github.io/jsx/#prod-HTMLHexCharacterReference
2. C1 Unicode control characters should not be replaced, `&#0;` should decode as U+0000, not U+FFFD, [this list](https://github.com/meriyah/meriyah/blob/a5cffce7523eda65efad607d7409e5b7b86c1518/src/lexer/decodeHTML.ts#L2134) shold be removed.
3. Should not decode codepoint>0x10FFFF. Same issue found in babel https://github.com/babel/babel/issues/18216

I'll fix these after #638

fisker:
Fixes #641

### DIFF
--- a/src/lexer/decodeHTML.ts
+++ b/src/lexer/decodeHTML.ts
@@ -264,7 +264,7 @@ const entities = new Map<string, string>(Object.entries({
 /* spellchecker: enable */
 
 export function decodeHTMLStrict(text: string): string {
-  return text.replaceAll(/&(?:[a-zA-Z]+|#x[\da-fA-F]+|#\d+);/g, (key) => {
+  return text.replaceAll(/&(?:[\da-zA-Z]+|#x[\da-fA-F]+|#\d+);/g, (key) => {
     if (key.charAt(1) === '#') {
       const secondChar = key.charAt(2);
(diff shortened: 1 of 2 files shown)


src/lexer/decodeHTML.ts
```diff
@@ -263,55 +263,14 @@ const entities = new Map<string, string>(Object.entries({
...
-  '155': 8250,
-  '156': 339,
-  '158': 382,
-  '159': 376,
-}).map(([key, value]): [number, number] => [Number(key), value]));
-
-// Used code from https://github.com/fb55/entities (BSD license)
 export function decodeHTMLStrict(text: string): string {
-  return text.replaceAll(/&(?:[a-zA-Z]+|#[xX][\da-fA-F]+|#\d+);/g, (key) => {
+  return text.replaceAll(/&(?:[\da-zA-Z]+|#x[\da-fA-F]+|#\d+);/g, (key) => {
```
fisker, line 267: There is a named entry with digit `there4` https://github.com/meriyah/meriyah/blob/9d237a27ee6dee9046bb648a6911a27635b52b25/src/lexer/decodeHTML.ts#L235

3cp (approved):
Thx!

PR #645, opened 2026-09-03
fix(jsx): count line terminators in JSXText

abhijeet117:
`nextJSXToken` counted JSXText line breaks incorrectly. It used mismatched character flags, so CRLF was counted twice, LFCR was counted once, and Unicode line separators were missed. The fix checks CR, LF, LS, and PS directly, treats CRLF as one terminator, normalizes CRLF to LF in JSXText values, and adds regression coverage for each case.

### DIFF
--- a/test/parser/miscellaneous/jsx.ts
+++ b/test/parser/miscellaneous/jsx.ts
@@ -64,7 +64,7 @@ describe('Miscellaneous - JSX', () => {
   // one (instead of two), and skipped `<LS>` / `<PS>` entirely. The decoded
   // `JSXText.value` also kept `<CR><LF>` raw instead of normalizing it to
   // `<LF>`, which Babel and acorn-jsx both do.
-  for (const [name, source, wantValue, wantEndLine] of [
+  for (const [name, source, expectedValue, expectedEndLine] of [
     ['<LF>', '<a>foo\nbar</a>', 'foo\nbar', 2],
     ['<CR>', '<a>foo\rbar</a>', 'foo\rbar', 2],
     ['<CR><LF>', '<a>foo\r\nbar</a>', 'foo\nbar', 2],
--- a/test/parser/miscellaneous/jsx.ts
+++ b/test/parser/miscellaneous/jsx.ts
@@ -57,6 +57,41 @@ describe('Miscellaneous - JSX', () => {
     });
   }
 
+  // `nextJSXToken` previously looked up line terminators through `CharTypes`,
+  // which has the `CarriageReturn`/`LineFeed` flags swapped relative to the
+  // conventional `\r`/`\n` mapping and only covers the first 128 code points.
+  // That made `<CR><LF>` two line terminators (instead of one), `<LF><CR>`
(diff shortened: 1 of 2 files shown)


3cp:
Looks like it's a follow up of #635. LGTM, thx!

I will wait for @morgan-coded and @fisker to have a look.

3cp (approved):

test/parser/miscellaneous/jsx.ts
```diff
@@ -57,6 +57,41 @@ describe('Miscellaneous - JSX', () => {
...
   }
 
+  // `nextJSXToken` previously looked up line terminators through `CharTypes`,
+  // which has the `CarriageReturn`/`LineFeed` flags swapped relative to the
+  // conventional `\r`/`\n` mapping and only covers the first 128 code points.
+  // That made `<CR><LF>` two line terminators (instead of one), `<LF><CR>`
+  // one (instead of two), and skipped `<LS>` / `<PS>` entirely. The decoded
+  // `JSXText.value` also kept `<CR><LF>` raw instead of normalizing it to
+  // `<LF>`, which Babel and acorn-jsx both do.
+  for (const [name, source, wantValue, wantEndLine] of [
```
fisker: `wantEndLine` -> `expectedEndLine`

fisker: `wantValue` -> `expectedValue`

morgan-coded (approved):
Checked the JSXText terminator paths against acorn-jsx for CRLF, LFCR, LS/PS, a trailing CR, and the following token locations; the values and locations matched in both script and module goals. The scanner now treats CRLF as one sequence while preserving lone CR and both Unicode separators, and the focused, plain, and test262 suites passed. Found no behavioral issue.

test/parser/miscellaneous/jsx.ts
3cp:
```suggestion
  for (const [name, source, expectedValue, expectedEndLine] of [
```

test/parser/miscellaneous/jsx.ts
```diff
@@ -57,6 +57,41 @@ describe('Miscellaneous - JSX', () => {
...
+    ['<PS>', '<a>foo\u2029bar</a>', 'foo\u2029bar', 2],
+    ['two <CR><LF>', '<a>\r\n\r\n</a>', '\n\n', 3],
+  ] as const) {
+    it(`${name} in JSX text counts as one line terminator`, () => {
+      const ast = parseSource(source, { jsx: true, loc: true });
+      const element = (ast.body[0] as ESTree.ExpressionStatement).expression as ESTree.JSXElement;
+      const text = element.children[0] as ESTree.JSXText;
+
+      t.equal(text.value, wantValue);
+      t.equal(text.loc?.end.line, wantEndLine);
```
3cp:
```suggestion
      t.equal(text.value, expectedValue);
      t.equal(text.loc?.end.line, expectedEndLine);
```

3cp (changes requested):

3cp (approved):

3cp:
v7.3.3 is released.

PR #646, opened 2026-09-07
fix(parser): restrict continue targets to iteration statement labels

MaxFreedomPollard:
`a: { for (;;) { continue a; } }` parses. A `continue` may only target a label of an iteration statement it is nested in, and `a` here labels a block, so ContainsUndefinedContinueTarget makes it an early SyntaxError. The same hole accepts a label on an `if`, on a `switch` case, on a `try` block, and on a block nested inside a loop. V8 and Acorn reject all of them, and the whitelisted test262 fixture `language/statements/block/labeled-continue.js` is exactly this shape.

`isValidLabel` in `src/common.ts` walks the label set chain outwards from the `continue` and cleared its "must be inside an iteration statement" requirement at the first set marked `loop`, then accepted the label wherever it turned up further out. Any loop between the `continue` and the label satisfied the check, whatever statement the label was actually attached to.

`parseIterationStatementBody` marks an iteration statement's body set with `loop`, and a labelled statement declares its label into the set it is itself parsed with, so the set directly above a `loop` set holds that iteration statement's own labels. The walk now tracks whether the set it is looking at is such a set, and only accepts a `continue` label declared there. `break` passes `isIterationStatement` as 0 and is untouched, so `a: { for (;;) { break a; } }` still parses; there is a test for that.

Verification on Node 22 against `main` at 56416fa. The seven newly rejected shapes fail on unmodified `main` with "Expect a ParserError thrown" and pass with the change; six accepted shapes are added alongside them, including `a: b: for (;;) { continue a; }`, `a: for (;;) { for (;;) continue a; }` and `a: while (x) { b: { continue a; } }`. `vitest run` is 91860 passing over 139 files with no existing snapshot touched. Full test262 at the pinned commit 7710052 goes from 8694 to 8696 invalid programs rejected while valid programs accepted stays at 94569, so nothing valid was tightened out, and the two `labeled-continue.js` whitelist lines come out. The ast alignment, jsx alignment and production suites pass, and `eslint`, `prettier --check`, `tsc`, `cspell`, `knip` and `generate-unicode.mjs --check` are clean.

### DIFF
--- a/test/parser/statements/__snapshots__/continue.ts.snap
+++ b/test/parser/statements/__snapshots__/continue.ts.snap
@@ -51,45 +51,45 @@ y++;
 `;
 
 exports[`Statements - Continue > Declarations - Continue > a: { b: for (;;) { continue a; } } 1`] = `
-"SyntaxError [1:29-1:30]: continue statement must be nested within an iteration statement
+"SyntaxError [1:28-1:29]: Label 'a' does not denote an iteration statement
 > 1 | a: { b: for (;;) { continue a; } }
-    |                              ^ continue statement must be nested within an iteration statement"
+    |                             ^ Label 'a' does not denote an iteration statement"
 `;
 
 exports[`Statements - Continue > Declarations - Continue > a: { for (;;) { continue a; } } 1`] = `
-"SyntaxError [1:26-1:27]: continue statement must be nested within an iteration statement
+"SyntaxError [1:25-1:26]: Label 'a' does not denote an iteration statement
 > 1 | a: { for (;;) { continue a; } }
-    |                           ^ continue statement must be nested within an iteration statement"
+    |                          ^ Label 'a' does not denote an iteration statement"
 `;
 
 exports[`Statements - Continue > Declarations - Continue > a: for (;;) { b: { for (;;) continue b; } } 1`] = `
-"SyntaxError [1:38-1:39]: continue statement must be nested within an iteration statement
+"SyntaxError [1:37-1:38]: Label 'b' does not denote an iteration statement
 > 1 | a: for (;;) { b: { for (;;) continue b; } }
-    |                                       ^ continue statement must be nested within an iteration statement"
+    |                                      ^ Label 'b' does not denote an iteration statement"
 `;
 
 exports[`Statements - Continue > Declarations - Continue > a: if (x) for (;;) { continue a; } 1`] = `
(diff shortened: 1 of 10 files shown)


test/parser/statements/__snapshots__/continue.ts.snap
```diff
@@ -50,6 +50,42 @@ y++;
   5 | } while(0);"
 `;
 
+exports[`Statements - Continue > Declarations - Continue > a: { b: for (;;) { continue a; } } 1`] = `
+"SyntaxError [1:29-1:30]: continue statement must be nested within an iteration statement
+> 1 | a: { b: for (;;) { continue a; } }
+    |                              ^ continue statement must be nested within an iteration statement"
```
3cp: The error location can be fixed by swapping `parseIdentifer` and `if (!isValidLabel)` in `src/parser.ts`.

There are two usages in `src/parser.ts`. Take one as example:

```ts
// one in parseContinueStatement
    if (!isValidLabel(parser, labels, tokenValue, /* requireIterationStatement */ 1))
      parser.report(Errors.UnknownLabel, tokenValue);
    label = parseIdentifier(parser, context | Context.AllowRegExp);
```

3cp: Also the printed error "continue statement must be nested within an iteration statement" sounds very confusing.

@fisker any suggestion to improve that error message? It probably should reflect the label was not for a loop statement.

Acorn uses "Unsyntactic break/continue" for all types of error. It isn't much better.

3cp (changes requested):
Thx! Only might need two small adjustments.

3cp:
@fisker we probably need to improve the type def for `export type Labels = any;`.

test/parser/statements/__snapshots__/continue.ts.snap
fisker: > Acorn uses "Unsyntactic break/continue" for all types of error. It isn't much better.

Agree. And  babel also use the same, I guess because they started as a fork of acorn.


Oxc uses

```
Label 'a' does not denote an iteration statement
```

better?

3cp: Definitely better than ours :-)

3cp (approved):

3cp:
Thank you!

PR #647, opened 2026-09-09
fix(parser): reject escaped `await` in async arrow parameters

morgan-coded:
In script code, an escaped `await` was accepted in async arrow parameters and defaults where the plain spelling is rejected.

The parser now matches the token by type and routes escaped `await` through the await-as-identifier checks, while the existing accepted script-identifier and non-async-arrow controls keep parsing.

The new rejection cases fail before the fix and pass with it; the parser suite and test262 pass with the whitelist unchanged.

This addresses the remaining case from #337; escaped async-function names and the pre-existing pattern gaps remain outside this change.


src/parser.ts
```diff
@@ -4520,6 +4524,7 @@ function parsePrimaryExpression(
   if ((parser.getToken() & Token.IsIdentifier) === Token.IsIdentifier) {
     switch (parser.getToken()) {
       case Token.AwaitKeyword:
+      case Token.AwaitKeyword | Token.IsEscaped:
```
3cp, line 4527: Just a question: does this mean the previous branch was unable to be reached before?

3cp (approved):

src/parser.ts
morgan-coded: Yes. In script code the lexer returns `Token.AwaitKeyword | Token.IsEscaped` for an escaped `await` (in modules and async functions it already returns `Token.EscapedReserved`), and the `switch` compares exact token values, so that token never matched `case Token.AwaitKeyword` and went down the plain-identifier path instead. That is why `parseAwaitExpressionOrIdentifier`, where the async-arrow parameter bookkeeping lives, never saw the escaped spelling; the new case only routes it there, and `isEscaped` then short-circuits to the identifier path since an escaped `await` cannot start an await expression.

PR #651, opened 2026-09-10
fix(parser): reject `await` in class field initializers

morgan-coded:
In modules, `class C { x = await 1 }` and classes inside concise async arrows accepted `await` expressions in field initializers.

Clearing `InGlobal` on entry to the initializer rejects those expressions while preserving `await` in computed keys, `extends` expressions and nested async arrows, and identifier use in scripts.

The new rejection cases fail before the fix and pass with it, the parser and test262 suites pass, and whitelist regeneration removes the two `await-identifier-module-3.js` lines.

This also allows valid object-literal accessor, async and generator method names written as escaped reserved words inside initializers, matching their existing behavior inside function bodies.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -8929,9 +8929,11 @@ function parsePropertyDefinition(
 
     if (parser.getToken() === Token.Arguments) parser.report(Errors.StrictEvalArguments);
 
+    // Initializer[+In, ~Yield, ~Await]: a module top-level `await` must not reach the initializer either.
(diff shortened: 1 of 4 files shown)


3cp (approved):
Thx! I am very surprised that `async function g() { return class { x = await }; }` is valid.

3cp:
Thanks for all the contributors. Our `test262/whitelist.txt` is now empty. 💯

PR #652, opened 2026-09-12
refactor: type `Labels` as a chained label set

### LINKED ISSUE #650: Improve the type def for `Labels`
> we probably need to improve the type def for `export type Labels = any;`.

_Originally posted by @3cp in https://github.com/meriyah/meriyah/issues/646#issuecomment-5576070592_

Yash121l:
`Labels` was `export type Labels = any`, so nothing in the label-set walk was checked. Neither the shape of the sets the parser builds nor the keys the helpers read off them. `isValidLabel` and `validateAndDeclareLabel` took `labels: any` on top of that.

I typed it as the object the parser actually builds. `$` links to the label set of the enclosing statement, `loop` is set only by `parseIterationStatementBody` to mark an iteration statement body's set, and a declared label is keyed by `$` followed by its name. The template literal index signature is the reason to prefer this over a plain `[key: string]` signature: it rejects keys that are not `$`-prefixed, which is the invariant the two helpers rely on.

The only runtime-adjacent change is that the three lookups index with a template literal instead of `'$' + name`, because a pattern index signature cannot be indexed by a plain `string`. The keys produced are identical, so there is no behaviour change and no snapshot moved.

I checked that the type fails on misuse rather than just passing: `labels.loop = 2`, `labels['plain' + name]`, `labels.$ = 1` and passing a string to `isValidLabel` each produce a compile error (TS2322, TS7053, TS2322, TS2345). All four compiled silently under `any`.

`vitest run` is 91949 passing over 139 files. `tsc`, `eslint`, `prettier --check`, `cspell` and `knip` are clean. The ast alignment and jsx alignment suites pass against test262 at `77100523`.

Fixes #650

### DIFF
--- a/src/estree.ts
+++ b/src/estree.ts
@@ -27,7 +27,10 @@ export interface Labels {
   $?: Labels;
   /** Marks the label set an iteration statement body is parsed with. */
   loop?: 1;
-  /** A label declared in this set, keyed by `$` followed by its name. */
+  /**
+   * A label declared in this set, keyed by `$` followed by its name. Always `1`;
+   * `Labels` is in the union only because the `$` key above matches this pattern.
+   */
   [label: `$${string}`]: Labels | 1 | undefined;
 }
 
--- a/src/estree.ts
+++ b/src/estree.ts
@@ -18,7 +18,18 @@ export interface Position {
   column: number;
 }
 
-export type Labels = any;
+/**
+ * The label set a statement is parsed with, chained outwards through `$` to the
+ * set of the statement it is nested in.
+ */
+export interface Labels {
+  /** The label set of the enclosing statement. */
+  $?: Labels;
+  /** Marks the label set an iteration statement body is parsed with. */
+  loop?: 1;
+  /** A label declared in this set, keyed by `$` followed by its name. */
+  [label: `$${string}`]: Labels | 1 | undefined;
+}
 
 export type IdentifierOrExpression = Identifier | Expression | ArrowFunctionExpression;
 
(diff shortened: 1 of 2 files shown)


src/estree.ts
```diff
@@ -18,7 +18,18 @@ export interface Position {
...
+ * The label set a statement is parsed with, chained outwards through `$` to the
+ * set of the statement it is nested in.
+ */
+export interface Labels {
+  /** The label set of the enclosing statement. */
+  $?: Labels;
+  /** Marks the label set an iteration statement body is parsed with. */
+  loop?: 1;
+  /** A label declared in this set, keyed by `$` followed by its name. */
+  [label: `$${string}`]: Labels | 1 | undefined;
```
3cp, line 34: Question: where is the `Labels` type for `labels["$name"]` being used? I couldn't find the usage.

Yash121l: It is not used. A label entry is always `1`. `Labels` is in that union only because the `$` key itself matches the `$${string}` pattern (empty string), and TypeScript requires every named property to be assignable to a matching index signature. With `1 | undefined` alone, `tsc` fails:

```
src/estree.ts:27:3 - error TS2411: Property '$' of type 'Labels | undefined' is not assignable to '`$${string}`' index type '1 | undefined'.
```

I have added that to the doc comment so the next reader does not have the same question. If you would rather avoid the overlap altogether, the alternative is to change the parent key at runtime to something that cannot collide with a label name, but that is a bigger change than this PR should carry.

3cp: Thx! I will merge this one then explore options to separate them and improve readability too.

3cp (approved):

src/estree.ts
Yash121l: Thanks. I will open a follow-up PR that separates the parent link from the label entries so the type no longer needs that union.

Yash121l: Opened #653 for the split.

PR #653, opened 2026-09-13
refactor: separate the label parent link from label entries

Yash121l:
Follow-up to #652, where @3cp said he would merge that one and then look at separating the two and improving readability.

In #652 `Labels` typed its entries as `Labels | 1 | undefined`, because the parent link lived in the same object as the declared labels under the key `$`, which matches the `` `$${string}` `` index signature. The union existed to accommodate one key, which made the type dishonest about what a label entry is.

This gives the parent link and the entries separate fields:

```ts
export interface Labels {
  parent?: Labels;
  loop?: 1;
  names?: Set<string>;
}
```

No union and no index signature. The `$` prefix on label names existed to keep names like `__proto__` and `toString` away from `Object.prototype`. A `Set` has no such hazard, so the prefix goes away too and the call sites read directly: `labels.names?.has(name)` and `labels = labels.parent`.

I also considered keeping the flat object and moving the parent to a `$$` key. That does not work: a label may be named `$`, so its entry key would be `$$` and collide with the parent. A `unique symbol` key avoids the collision but reads worse than a named field.

`names` is optional and created on the first declaration rather than at construction. Every block, switch case, loop body and try block builds a label set, and almost none of them declare a label, so the common path still allocates exactly one object literal and the five `{}` roots are unchanged.

There is no benchmark script in the repo, so I measured by hand: built `dist/meriyah.mjs` on `main` and on this branch and parsed `typescript.js` (about 9 MB), 7 runs per build, twice per side. Medians were within a few milliseconds of each other in both directions, inside run-to-run noise.

Behaviour is unchanged. Verified with `npm run lint:types`, `npx vitest run` (91949 tests, 139 files, all passing), `TEST262=1 npx vitest run test/conformance` for the ast-alignment, jsx-ast-alignment and test262-parser tests, and `npm run lint` (eslint, prettier, cspell, knip).

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -395,13 +395,13 @@ export function isValidLabel(
   let isIterationLabelSet: 0 | 1 = 0;
 
   while (labels) {
-    if (labels[`$${name}`]) {
+    if (labels.names?.has(name)) {
       // A label is declared at most once per chain, so this is its only declaration.
       if (isIterationStatement && !isIterationLabelSet) parser.report(Errors.InvalidNestedStatement, name);
       return 1;
     }
     isIterationLabelSet = labels.loop ? 1 : 0;
-    labels = labels.$;
+    labels = labels.parent;
   }
 
   return 0;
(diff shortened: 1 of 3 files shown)


3cp (approved):
Thx! Was thinking of about the same :)

PR #654, opened 2026-09-13
refactor: type callee and class field values in estree

Yash121l:
Three fields in `src/estree.ts` were typed `any`. `CallExpression.callee` even carried the intended type in a trailing comment. This replaces them with the types the spec and the parser already imply:

- `CallExpression.callee`: `any` to `Expression | Super` (comment removed)
- `PropertyDefinition.value`: `any` to `Expression | null`
- `AccessorProperty.value`: `any` to `Expression | null`

These match ESTree (es2015 `CallExpression.callee`, es2022 `PropertyDefinition.value` and `AccessorProperty.value`), and they match what the parser actually builds. I checked every construction site rather than trusting the spec alone.

One knock-on change in `src/parser.ts`: `parseAsyncArrowOrCallExpression` declared `callee: ESTree.Identifier | void`, but all three callers pass an `ESTree.Identifier`. The `| void` was never reachable and was the only thing the old `any` was hiding, so I dropped it. No `as any` added anywhere.

For the class fields, `parsePropertyDefinition` already declares its local as `ESTree.Expression | null`, so the new field types are exact, not just spec-shaped.

Verified with `npm run lint:types`, the full `vitest` suite under `CI=1` (91952 tests, including the AST alignment and test262 conformance runs), the production test after a build, plus eslint, prettier, cspell and knip. I also temporarily assigned a wrong node type at one call site and confirmed `tsc` rejects it, then reverted.

Same approach as #652.

### DIFF
--- a/src/estree.ts
+++ b/src/estree.ts
@@ -345,7 +345,7 @@ export interface ChainExpression extends _Node {
 
 export interface CallExpression extends _Node {
   type: 'CallExpression';
-  callee: any; //Expression | Super;
+  callee: Expression | Super;
   arguments: (Expression | SpreadElement)[];
   optional: boolean;
 }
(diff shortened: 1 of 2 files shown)


3cp (approved):
Thx!

PR #655, opened 2026-09-13
fix(jsx): compare namespaced tag names by text, and type the common.ts helpers

Yash121l:
Follow-up to #652 and #654. `src/common.ts` still had four helpers with `any` parameters or `any` return types. This replaces them with the types the functions actually work with, and typing one of them exposed a bug.

The bug: `isEqualTagName` handles `JSXNamespacedName` by concatenating `elementName.namespace + ':' + elementName.name`, which are two AST nodes, so every namespaced tag name stringified to `[object Object]:[object Object]`. All namespaced tag names therefore compared equal. `<a:b></c:d>` parsed without an error, and the message for `<path></svg:path>` was `Expected corresponding JSX closing tag for [object Object]:[object Object]`, which had been committed as a snapshot. It now recurses into `namespace` and reads `name.name`. I added `<a:b></c:d>` to the JSX fail list and the `<path></svg:path>` snapshot now says `svg:path`. With the switch exhaustive, the `default` branch and its `istanbul ignore next` are gone.

The typing changes:

- `reinterpretToPattern(parser, node)` takes `ESTree.Node` instead of `any`. It cannot take just the five node kinds its switch handles, because callers hand it a `for` statement's `init`, `parseAssignmentExpression`'s `left`, and arrow params, so it takes a node and narrows inside. It rewrites the node in place, and TypeScript will not let you assign a new `type` to a narrowed node, so the object is bound once through a small local view type where `type` is writable and `operator` is optional. No casts.
- `isPropertyWithPrivateFieldKey(expr)` takes `ESTree.Expression` and checks `expr.type === 'MemberExpression' && expr.property.type === 'PrivateIdentifier'`. Same behaviour: the only other node with a `property` that can reach it is `JSXMemberExpression`, whose `property` is a `JSXIdentifier`.
- `isEqualTagName(elementName)` takes `ESTree.JSXTagNameExpression` and returns `string`.
- `classifyIdentifier` returns `void`. It only reports and sets flags.

Two call sites in `src/parser.ts` pass values typed `... | null`. Both are non-null in practice: the first sits behind a guard that already dereferences `left.type`, and in the second every reachable branch assigns `argument` (the `null` only survives because the surrounding `parsePrimaryExpression` and `parseMemberOrUpdateExpression` still return `any`). I used a non-null assertion at both rather than widening the parameter to accept `null`, since the function dereferences `node.type` straight away.

Verified with `npm run lint:types`, the full `vitest run` including the test262 conformance and acorn alignment suites (91959 tests passing), `npm run build` plus the production tests, and eslint, prettier, cspell and knip.

### DIFF
--- a/test/parser/miscellaneous/jsx.ts
+++ b/test/parser/miscellaneous/jsx.ts
@@ -155,6 +155,8 @@ describe('Miscellaneous - JSX', () => {
     { code: '<:path />', options: { jsx: true } },
     { code: '<path></svg:path>', options: { jsx: true } },
     { code: '<a:b></c:d>', options: { jsx: true } },
+    { code: '<a:x></b:x>', options: { jsx: true } },
+    { code: '<x:a></x:b>', options: { jsx: true } },
     { code: '<foo.bar></foo.baz>', options: { jsx: true } },
     { code: '<chinese:alladinfoo.bar></foo.baz>', options: { jsx: true } },
     { code: '<foo:bar></foo.baz>', options: { jsx: true } },
--- a/test/parser/miscellaneous/jsx.ts
+++ b/test/parser/miscellaneous/jsx.ts
@@ -154,6 +154,7 @@ describe('Miscellaneous - JSX', () => {
     { code: '<f><g/></ff>;', options: { jsx: true } },
     { code: '<:path />', options: { jsx: true } },
     { code: '<path></svg:path>', options: { jsx: true } },
+    { code: '<a:b></c:d>', options: { jsx: true } },
     { code: '<foo.bar></foo.baz>', options: { jsx: true } },
     { code: '<chinese:alladinfoo.bar></foo.baz>', options: { jsx: true } },
     { code: '<foo:bar></foo.baz>', options: { jsx: true } },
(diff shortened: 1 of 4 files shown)


test/parser/miscellaneous/jsx.ts
```diff
@@ -154,6 +154,7 @@ describe('Miscellaneous - JSX', () => {
     { code: '<f><g/></ff>;', options: { jsx: true } },
     { code: '<:path />', options: { jsx: true } },
     { code: '<path></svg:path>', options: { jsx: true } },
+    { code: '<a:b></c:d>', options: { jsx: true } },
```
fisker, line 157: `<a:x></b:x>` `<x:a></x:b>` too?

Yash121l: Added both in ae309b8. `<a:x></b:x>` and `<x:a></x:b>` now fail with `Expected corresponding JSX closing tag for b:x` / `x:b`. On main both parse without error, since every namespaced name compared as `[object Object]:[object Object]`.

3cp (approved):

PR #656, opened 2026-09-13
fix(parser): report yield-in-parameter errors at the yield token

morgan-coded:
`yield` errors in arrow parameters inside generators pointed at `=>`, and generator-parameter errors pointed at the token after `yield`.

This records `firstYieldLocation` alongside the await location and anchors the generator-parameter check at the consumed token, preserving accept/reject behavior.

The eight new location assertions fail before the fix and pass with it; 68 existing snapshots move onto `yield`, and the parser suite and test262 pass.

Some parameter lists with a nested function, method, class or async arrow after the offending `yield` still report at `=>`, matching the existing limitation on the await side.


3cp (approved):
Thx!

Issue #199, opened 2021-12-10
TypeScript: meriyah package is not compatible with estree package


fis-cz:
Hello,

Just trying to use 'meriyah' (4.2.0) with 'estree-toolkit' (1.2.6) and '@types/estree' (0.0.50), so latest available versions, with TypeScript compiler. I just found that Meriyah is not compatible with '@types/estree'.

When I do:

```

import { Program } from "estree";
import { parseModule } from "meriyah";

const ast: Program = parseModule("");

```

the type checker reports the error:

`
Type 'import(".../node_modules/meriyah/dist/src/estree").Program' is not assignable to type 'import(".../node_modules/@types/estree/index").Program'.
  Types of property 'body' are incompatible.
    Type 'Statement[]' is not assignable to type '(Directive | Statement | ModuleDeclaration)[]'.
      Type 'Statement' is not assignable to type 'Directive | Statement | ModuleDeclaration'.
        Type 'BlockStatement' is not assignable to type 'Directive | Statement | ModuleDeclaration'.
          Type 'import(".../meriyah/dist/src/estree").BlockStatement' is not assignable to type 'import(".../node_modules/@types/estree/index").BlockStatement'.
            Types of property 'body' are incompatible.
              Type 'import(".../node_modules/meriyah/dist/src/estree").Statement[]' is not assignable to type 'import(".../node_modules/@types/estree/index").Statement[]'.
                Type 'import(".../node_modules/meriyah/dist/src/estree").Statement' is not assignable to type 'import(".../node_modules/@types/estree/index").Statement'.
`

Although I can go around it by:

```

import { Program } from "estree";
import { parseModule } from "meriyah";

const ast: Program = <any>parseModule("");

```

It would be great if this works out of the box.

3cp:
I am also using forced type conversion to tap into various estree tools.
```ts
import {parseScript, Options} from 'meriyah';
import * as ESTree from 'estree';

const paserOptions: Options = {
  next: true,
  ranges: true,
  loc: true,
  webcompat: true,
  globalReturn: true,
  impliedStrict: true
};

export default function (code: string): ESTree.Program {
  return parseScript(code, paserOptions) as ESTree.Program;
}
```

It's possible to refactor meriyah code to use estree types (that will introduce a dependency which we don't like), but needs quite some work.

[...]

PR #657, opened 2026-09-13
test(estree): assert node type assignability to estree

Yash121l:
Ref #199.

#199 tracks meriyah's AST types not lining up with the `estree` package. Before changing any of those types I wanted a way to see exactly where they diverge and keep that visible as it gets fixed, so this adds a type-only test.

`test/estree-compat.ts` takes every node type meriyah and estree have in common, checks whether meriyah's version is assignable to estree's, and pins the ones that are not into a `KnownGaps` list. The check is an exact equality, so tsc fails both when a gap closes (remove the name from the list) and when a new one appears. There are no runtime assertions. `npm run lint:types` compiles the file, and it is excluded from vitest's `test/**/*.ts` glob the same way `test/test-utils.ts` is. `@types/estree` and `@types/estree-jsx` are added as devDependencies.

65 of the 87 shared node types are not assignable today, including `Program`. The 22 that are: `BreakStatement`, `ContinueStatement`, `DebuggerStatement`, `EmptyStatement`, `ExportAllDeclaration`, `ExportSpecifier`, `Identifier`, `ImportAttribute`, `ImportDeclaration`, `ImportDefaultSpecifier`, `ImportNamespaceSpecifier`, `ImportSpecifier`, `JSXClosingFragment`, `JSXEmptyExpression`, `JSXIdentifier`, `JSXOpeningFragment`, `Literal`, `MetaProperty`, `PrivateIdentifier`, `Super`, `TemplateElement`, `ThisExpression`. `AccessorProperty`, `Decorator` and `ParenthesizedExpression` have no estree counterpart at all and are pinned separately.

Following the errors, most of the 65 come from one place. meriyah's `Expression` union is wider than estree's: it includes `RestElement`, `SpreadElement`, the JSX opening and closing nodes, `JSXSpreadChild`, and `ClassDeclaration` via `LeftHandSideExpression`. Every node with an `Expression`-typed child inherits that. The independent ones are `AssignmentExpression`, `BinaryExpression` and `LogicalExpression` (`operator: string` rather than estree's literal unions), `JSXText` (`raw`), and `Property` (estree splits `Property` and `AssignmentProperty`). I will narrow `Expression` in a follow-up PR, which should close most of the list at once.

On the `start`/`end` fields from the first item of your list in #199: those are already there. `_Node` in `src/estree.ts` declares `start?: number` and `end?: number`, and `finishNode` sets them when `ranges` is enabled. So nothing changed for that. The `Comment` type values and `leadingComments`/`trailingComments` are still open and out of scope here.

Verified with `npm run lint:types`, `npx vitest run` (91952 passing), eslint, prettier, cspell and knip, all clean. I also checked the two-way behaviour by hand: removing a name from `KnownGaps` and adding a name that already passes both make tsc fail.


3cp (approved):

3cp:
LGTM!
@fisker please have a look.

fisker (approved):

Yash121l:
@3cp are you happy for me to narrow the `Expression` union now? That was the one change that closed most of the 65 gaps the type test pins, and I'd rather do it as its own PR against the test so the diff shows exactly which assignability cases flip.

3cp:
Yes please.

PR #661, opened 2026-09-18
refactor(estree): narrow the Expression union to expression nodes

Yash121l:
Follow-up to #657, which @3cp gave the go-ahead for there.

`Expression` listed a number of node types that are not expressions: `RestElement`, `SpreadElement`, the JSX opening and closing boundary nodes, `JSXExpressionContainer`, `JSXSpreadChild`, and through `LeftHandSideExpression` and `PrimaryExpression` also `ClassDeclaration`, `ArrayPattern`, `ObjectPattern` and `Super`. Every node with an `Expression`-typed child inherited all of them, so the types admitted trees the parser never produces.

`Expression` now holds exactly the node types estree's `Expression` does: its 25 members plus `JSXElement` and `JSXFragment` from estree-jsx. Both directions of the difference are empty. `NewExpression.arguments` gains `SpreadElement`, which the wide union used to supply and which estree declares too.

**I have to correct something I said in #657.** I predicted this would close most of the 65 gaps the type test pins. It closes none of them, and I checked with a per-type probe rather than assuming. The gap count is 65 before and 65 after, so `test/estree-compat.ts` is untouched here.

The reason is that the assignability graph is one mutually recursive cycle: `ExpressionStatement.expression` to `Expression` to `ArrowFunctionExpression.body` and back to `Expression`. `Expression` cannot become assignable until all 27 of its members are, and several have independent defects. `AssignmentExpression.left` is `Expression` where estree has `Pattern | MemberExpression`, `operator` is `string` on `AssignmentExpression`, `BinaryExpression` and `LogicalExpression`, `ArrayPattern.elements` is `Expression[]`, and `Property` stands where estree splits out `AssignmentProperty`. I tried tightening the three operator types on top of this change and that flips nothing either. The gaps will move in one batch once the defects inside the cycle are fixed together, not one union at a time.

So this is a prerequisite rather than the self-demonstrating diff I had in mind. I am happy to park it if you would rather see one combined PR that fixes the cycle and flips the list in a single visible step. The narrowing is still correct on its own, but I would understand not wanting it landed for no movement on the list.

The parser had been using `Expression` as its cover-grammar type. The places that hold a node before deciding whether it is an expression or a binding pattern now say so, and the internal signatures that threaded `ArgumentExpression` only to pick up `SpreadElement` take `Expression`. I kept to the `as ESTree.Expression` idiom already used on neighbouring lines and added no `as any`.

This is breaking for TypeScript consumers even though runtime is untouched. Narrowing `Expression` breaks input positions that used to accept spread, rest or patterns, and widening `NewExpression.arguments` means readers must handle `SpreadElement`. `ArgumentExpression` is now unused inside `src/`, but I left it exported rather than add a second breaking removal.

Verified with `npm run lint` (eslint, tsc, prettier, cspell, knip and unicode all clean) and `npx vitest run` (91986 passing, unchanged from main). Type only: `dist/meriyah.min.mjs` is byte identical before and after, and the unminified bundles differ only in formatting.


3cp (approved):

3cp:
Thx!
The breaking change for type is fine. We will release a major version after all were finished.

3cp:
@fisker please have a look too.

Yash121l:
Thanks. I'll take the rest of the cycle once this lands, in one PR so the 65 gaps move in a single visible step: `AssignmentExpression.left` to `Pattern | MemberExpression`, literal operator unions on `AssignmentExpression`, `BinaryExpression` and `LogicalExpression`, `ArrayPattern.elements` off `Expression[]`, and splitting `AssignmentProperty` out of `Property`. That one will update `test/estree-compat.ts`, which is where the payoff finally shows.

fisker (approved):
I don't have much experience on TypeScript, but LGTM!

PR #662, opened 2026-09-18
refactor(estree): type binding positions as patterns and pin operator unions

Yash121l:
Follow-up to #661, which @3cp approved with "the breaking change for type is fine, we will release a major version after all were finished".

I said on #661 that this one would move the 65 gaps `test/estree-compat.ts` pins in a single visible step. **It moves three.** I measured with a per-type probe rather than assuming, and I want to be plain about the distance between what I promised and what this delivers.

Here is what changed. `AssignmentExpression.left` was `Expression` where estree has `Pattern | MemberExpression`. `ArrayPattern.elements` was `Expression[]` where estree has `(Pattern | null)[]`. `RestElement.argument`, `SpreadElement.argument`, `VariableDeclarator.id` and the for-in and for-of `left` had the same shape of defect, the parser's cover grammar leaking into the published types, so I fixed those in the same pass. `AssignmentExpression`, `BinaryExpression` and `LogicalExpression` now carry estree's literal operator unions instead of `string`, and `AssignmentProperty` is split out of `Property` as estree does.

That is everything I listed on #661, and on its own it closes **zero** gaps. The three that do close come from a separate defect I found while chasing the cycle: `JSXNamespacedName.namespace` and `JSXMemberExpression.object` accepted node types the JSX grammar never produces there. `parseJSXNamespacedName` is only ever reached with an identifier, and `parseJSXMemberExpression` already took the narrow type. Narrowing both closes `JSXNamespacedName`, `JSXMemberExpression` and `JSXClosingElement`.

**Why the rest cannot close.** `ClassBody.body` contains `AccessorProperty`, and `@types/estree` has no `AccessorProperty` node at all. `ClassExpression` is a member of `Expression`, so `Expression` is not assignable to estree's `Expression`, and every one of the remaining 62 gaps has an `Expression`-typed field somewhere. The test already pins `AccessorProperty` as a node type missing from estree; what I did not see when writing #657 and #661 is that a missing node type sitting inside `ClassBody` also blocks everything reachable from `Expression`. Nothing meriyah can do to its own types fixes that. It needs `AccessorProperty` in `@types/estree`, or for meriyah to stop threading it into `ClassBody`. So the cycle is not one PR away from closing, it is blocked on DefinitelyTyped. I am happy to send that PR upstream if you think the node belongs there.

To find that I ran the type test with meriyah's `Expression`, `Statement` and `Pattern` replaced by estree's own, which leaves only genuine per-interface defects visible. That count went from 34 to 18 with this change. The 18 that remain, if you want the map: `ClassBody` and `AccessorProperty`, not fixable here; `FunctionDeclaration.id` and `ClassDeclaration.id` nullable, which needs estree's `MaybeNamed*` split; `AssignmentPattern.right`, `FunctionExpression.body` and `MethodDefinition.key` optional or nullable; `ObjectExpression.properties` admitting `MethodDefinition` and `RestElement`; the two export declarations; `JSXAttribute.value` and `JSXSpreadChild.expression`; and `JSXText.raw`, which is only emitted under `options.raw` and so needs a runtime decision rather than a type one. Separately, the `Statement` union still contains `ClassExpression` and the export declarations where estree's does not, which is another #661-shaped narrowing that I would rather do on its own.

I kept to the `as ESTree.X` idiom on the five parser sites this touches and added no `as any`. `PropertyName` is now referenced by no interface; I left it exported rather than add a second breaking removal, the same call I made for `ArgumentExpression` on #661.

Breaking for TypeScript consumers, not at runtime. The input positions listed above all narrow, so constructing nodes from a wider `Expression` stops compiling, and `.operator` stops accepting an arbitrary `string`. `SpreadArgument` is removed; `AssignmentOperator`, `BinaryOperator`, `LogicalOperator` and `AssignmentProperty` are added. The one widening is `ArrayPattern.elements`, which now admits `null` for holes, so readers must handle them.

Verified with `npm run lint` (eslint, tsc, prettier, cspell, knip and unicode all clean) and `npx vitest run` (91986 passing, unchanged from main). Type only: `dist/meriyah.min.mjs`, `dist/meriyah.cjs`, `dist/meriyah.umd.min.js` and `dist/meriyah.mjs` are all byte identical before and after.


src/estree.ts
```diff
@@ -281,10 +281,13 @@ export interface ArrowFunctionExpression extends _Node {
...
 
+export type AssignmentOperator =
+  '=' | '+=' | '-=' | '*=' | '/=' | '%=' | '**=' | '<<=' | '>>=' | '>>>=' | '|=' | '^=' | '&=' | '||=' | '&&=' | '??=';
+
 export interface AssignmentExpression extends _Node {
   type: 'AssignmentExpression';
-  operator: string;
-  left: Expression;
+  operator: AssignmentOperator;
+  left: Pattern | MemberExpression;
```
fisker: Missing `CallExpression`?

Yash121l: Yes, and in three more places than one. With `webcompat: true` meriyah parses all of these and puts a `CallExpression` in the target:

```js
f() = 1                 // AssignmentExpression.left
[...f()] = []           // RestElement.argument
for (f() in {}) ;       // ForInStatement.left
for (f() of []) ;       // ForOfStatement.left
```

Fixed in 7af802d by admitting `CallExpression` in those four positions. estree does not have it there, so those types stay in the gap list, but the point of these types is to describe what the parser actually produces and I had just broken that. Good catch, thank you. Bundles are still byte identical and the corpus is still 91986 passing.

fisker: `a()++`

Yash121l: That one is already right, so I left it alone. `UpdateExpression.argument` is `Expression`, which includes `CallExpression`, and this PR does not narrow it. `a()++`, `++a()` and `a()--` all parse under `webcompat: true` and give `UpdateExpression -> CallExpression`, which the current type accepts; estree types that field as `Expression` too, so narrowing it would diverge for no gain. `delete a()` is the same story and parses without webcompat.

3cp (approved):

3cp:
Let me know if it's ready for merge.

Yash121l:
Ready from my side. fisker's catch is fixed in 7af802d, the bundles are byte identical and the corpus is at 91986 passing.

fisker (approved):

